PR #4, opened 2022-10-18
Feature/bc 300 implement create project

mendo94:
Tried to get this done asap because this ticket is blocking a lot of other tickets, plus I know you will have much for me to change/fix. Was not too fixated on styling - just getting the api call to work and the schema to match the user stories. I was only focused on getting the project to post to the database - NOT the roles because I know that is coming in a later ticket.


src/screens/CreateProject/CreateProject.tsx
```diff
@@ -1,9 +1,161 @@
...
+  project?: CreateProject;
+}
+
+const CreateProject: React.FC<Props> = ({ project }) => {
+  const [fileSelected, setFileSelected] = useState<File>();
+  const [createProjects, setCreateProjects] = useState<any[]>([]);
+  const navigate = useNavigate();
+
+  const handleProjectInputChange = (
+    e: React.ChangeEvent<HTMLInputElement | HTMLSelectElement | HTMLTextAreaElement>,
```
yeezick, line 37: good idea

src/screens/CreateProject/CreateProject.tsx
```diff
@@ -1,9 +1,161 @@
...
 
-const CreateProject: React.FC<CreateProjectProps> = (props) => {
-  return <div>CreateProject</div>;
+interface Props {
+  project?: CreateProject;
+}
+
+const CreateProject: React.FC<Props> = ({ project }) => {
+  const [fileSelected, setFileSelected] = useState<File>();
+  const [createProjects, setCreateProjects] = useState<any[]>([]);
```
yeezick, line 33: we should find out the specific type for the `createProjects` array 
also curious about the `File` type, where's it coming from?

src/screens/CreateProject/CreateProject.tsx
```diff
@@ -1,9 +1,161 @@
...
+          name="overview"
+          className="overview"
+          cols={30}
+          rows={10}
+          onChange={(e) => handleProjectInputChange(e)}
+        ></textarea>
+
+        <p>Save project as a draft or publish?</p>
+        <div className="checkbox-container">
+          <label htmlFor="status">
```
yeezick, line 129: i think we should discuss as a team about the best way to create label/input pairs. above you wrote them as:
```
<label>i am label</label>
<input />
```
and here they're written as 
```
<label>
    i am label
    <input />
</label>
```

I personally prefer the first format because it gives us control of the input & label separately, however the second format eliminates the need for the "htmlFor" and the "id" attributes ("id" bubbles up to the global scope of the browser html and is an opportunity for clashes if there are similar id's on other files

yeezick: Also, I think this div would call for a radio button over checkbox

src/screens/CreateProject/CreateProject.tsx
```diff
@@ -1,9 +1,161 @@
...
+      <p className="heading">Create a Project</p>
+
+      <form onSubmit={handleNewProject} className="project-form" noValidate>
+        <div className="photo-container">
+          <label htmlFor="photo" className="photo">
+            <RiUploadCloudFill size={25} />
+            Upload Image
+            <input accept="image/*" id="photo" name="photo" type="file" multiple={false} onChange={handleImageChange} />
+          </label>
+        </div>
```
yeezick, line 76: add some text below this input in red font to let us know this feature is not built quite yet, but we can keep it on the gui

src/screens/CreateProject/CreateProject.tsx
yeezick: once you git pull with development, you should have the Project Interface in `/src/utilities/Interfaces/ProjectInterface.ts`

yeezick (changes requested):
resolve conversations above, ping me to test, and should be ready to merge!

src/screens/CreateProject/CreateProject.tsx
yeezick: let's also rename `createProjects` to `projectForm` and set the initial state to an empty or dummy object instead

src/screens/CreateProject/CreateProject.tsx
```diff
@@ -1,9 +1,161 @@
...
+
+        <label htmlFor="title">Title</label>
+        <input type="text" name="title" onChange={handleProjectInputChange} />
+
+        {/* Will be able to remove this section once a user is able to be logged in and token or userID is stored in redux store or local storage. Must enter the id in this label until then or api call will fail. */}
+        <label htmlFor="project-owner">Project Owner</label>
```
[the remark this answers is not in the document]
mendo94: Waiting to test this out better once the sign in screen is built

src/screens/CreateProject/CreateProject.tsx
```diff
@@ -1,9 +1,161 @@
...
+        <input type="text" name="project_owner" onChange={handleProjectInputChange} />
+
+        <label htmlFor="technologies_used">Technologies Used (separate by commas)</label>
+        <input
+          list="technologies"
+          type="email"
```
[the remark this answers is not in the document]
mendo94: from yesterday's convo, yes, it returns a string

src/screens/CreateProject/CreateProject.tsx
mendo94: Changed to radio button and moved the labels to <label>i am label</label>
<input /> since I do not need the checkbox here.

src/screens/CreateProject/CreateProject.tsx
mendo94: red text added here

src/screens/CreateProject/CreateProject.tsx
yeezick: This conversation was not resolved.

yeezick:
@mendo94 one last item to resolve. also, i noticed the email on the project details page is camelcased, is it returning like that from the api?

src/screens/CreateProject/CreateProject.tsx
mendo94: This was resolved, not sure why it didn't show up here that it was changed.

mendo94:
It was just some styling getting passed from a different component. I removed it - but it was not returning like that from the api to answer your question.

mendo94:
Changed project interface so the useState could pass the data coming from the api and took out. the empty array. Needed to refactor all of the project components due to this change. Ready to be reviewed again.

src/screens/CreateProject/CreateProject.tsx
```diff
@@ -8,27 +8,76 @@ import './CreateProject.scss';
...
+        },
+        technologies_used: [],
+        title: '',
+        updatedAt: '',
       });
     } else {
       navigate('/sign-in');
     }
-  };
+  }, [setProjectForm]);
```
yeezick, line 75: since setProjectForm is the setter function, it's value would never change which is why this works but i'm not sure it's best practice... 
do we need anything in the dependency array at all?

src/screens/Projects/ProjectDetails.tsx
```diff
@@ -3,22 +3,36 @@ import { getOneProject } from '../../utilities/api/projects';
...
+}
+
 const ProjectDetails: React.FC = (): JSX.Element => {
   const [project, setProject] = useState<ProjectInterface>({
-    id: new String(),
-    title: new String(),
-    duration: new String(),
-    meeting_cadence: new String(),
-    overview: new String(),
+    id: '',
```
yeezick, line 12: i see now about having to recreate to rewrite this object so many times... sorry!

If I could suggest creating a js file at `src/utilities/constants.js` and export this object as a variable from there, we would make the dummy object easier to maintain by only having 1 place to update it as well as make this file easier on the eyes

mendo94: Cleaned up the project files by moving the dummy object into a constants js file

src/screens/CreateProject/CreateProject.tsx
mendo94: took out dependency

yeezick (dismissed):
excellent work rina! this is ready to go!

yeezick (approved):

PR #6, opened 2022-10-19
Feature/bc 323 sign in functionality

zendrias:
- Created login screen.
- Check username and password fields before submit button is enabled.
- Handle 201 HTTP response as well as 401 response gracefully. ( Code 401 API response was throwing Axios error )
- Rendering user/unauthenticated-user JSON to the screen.
- Remove the token from local storage if the user submits invalid credentials while logged in.


src/api/apiConfig.js
```diff
@@ -23,3 +23,10 @@ api.interceptors.request.use(
     return Promise.reject(error);
   }
 );
+
+
+api.interceptors.response.use(response => {
+  return response;
+}, error => {
+  return error.response;
+});
```
yeezick, line 32: curious about what this does

src/api/users.js
```diff
@@ -59,6 +59,10 @@ export const signUp = async (credentials) => {
 export const signIn = async (credentials) => {
   try {
     const res = await api.post('/sign-in', credentials);
+    if (res.status === 401) {
+      if (localStorage.getItem('token')) localStorage.removeItem('token');
+      return res.data
+    }
```
yeezick, line 65: this too, how did the user receive unauthorized but had a token?
it should never be the case anyway, we haven't blocked that user flow yet but a user should never a see a sign-in page if they're already logged in

src/screens/SignIn/SignIn.tsx
```diff
@@ -1,7 +1,58 @@
...
+          ref={passwordRef}
+          required
+        />
+
+        <button disabled={buttonDisabled} type="submit">Go</button>
+
+      </form>
+      <pre>{JSON.stringify(apiResponse, null, 2)}</pre>
+    </>
+  );
```
yeezick, line 55: would love to hear about why you chose refs over inputs & state, i've never seen this method for forms before

src/screens/SignIn/SignIn.tsx
```diff
@@ -1,7 +1,58 @@
...
+    <>
+      <form onSubmit={handleSubmitForm}>
+
+        <input
+          type="email"
+          onChange={handleFormDataChange}
+          ref={emailRef}
+          required
+        />
+
```
yeezick, line 42: please add labels & style forms with flex columns for readability

yeezick (changes requested):
just a few questions and a suggestion

the api also fails with:
`"Cannot read properties of null (reading '_doc')"`
if you made changes to the backend, there's no pr for it 

now that Koffi has integrated the redux store, this would be good to merge with development and store the authenticated user in redux. you can also route the user to `/users/:id`, I should be done with UserProfile by the time this gets merged

src/api/apiConfig.js
zendrias: From what I've seen so far using Axios, if Axios gets an HTTP response code, other than 200 - 299, it is automatically classified as an error. So the 401 response was throwing an error.  so I'm catching the "error", and returning just the unauthorized message here.  That's what worked out for me.

src/api/users.js
zendrias: I thought I might as well add it in case, but touche.

src/screens/SignIn/SignIn.tsx
zendrias: I think it looks cleaner, I got to use a less common hook, and I saw no reason for needing state. I'm open to new ideas though!

yeezick:
@zendrias so I looked deeper into refs vs. state, mainly [this article](https://css-tricks.com/react-forms-using-refs/) and while interesting, i don't think this would be the right use case for refs. Similarly, this would be the only form using refs or state in the app and it might be better to stick to one method until we have a reason to use an alternative.
I admire the initiative though! Made me want to look deeper into refs.

Once you've handled a conversation above, resolve them so i know they're taken care of.

src/layout/Nav/Nav.tsx
```diff
@@ -8,6 +8,8 @@ export const Nav: React.FC = () => {
   const [authLinks, setAuthLinks] = useState<boolean>(false);
   const authUser = useSelector(selectAuthUser);
 
+  console.log(authUser)
```
yeezick, line 11: remove console log

src/screens/SignIn/SignIn.module.css
```diff
@@ -0,0 +1,64 @@
...
+}
+
+.sign_in_container {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  height: calc(100vh - 80px);
+}
+
```
yeezick, line 14: whitespace

src/screens/SignIn/SignIn.module.css
```diff
@@ -0,0 +1,64 @@
+* {
+  margin: 0;
```
yeezick, line 2: is this necessary? i saw you applied to index.scss as well, doesn't that apply to all children of the index.ts file?

src/screens/SignIn/SignIn.module.css
```diff
@@ -0,0 +1,64 @@
...
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  height: calc(100vh - 80px);
+}
+
+
+.sign_in_form {
+  width: 40vmin;
```
yeezick, line 17: please alphabetize

src/screens/SignIn/SignIn.tsx
```diff
@@ -1,5 +1,88 @@
...
+import { useDispatch } from 'react-redux';
+import { AppDispatch } from '../../utilities/redux/store'
+import { setAuthUser } from '../../utilities/redux/slices/users/userSlice'
+import { useNavigate } from 'react-router-dom';
 
-export const SignIn: React.FC = () => {
-  return <div>SignIn</div>;
-};
\ No newline at end of file
+interface FormData {
```
yeezick, line 9: move this to `/utilities/types/UserInterface` and rename to something more semantic like `SignInInterface`

src/screens/SignIn/SignIn.tsx
```diff
@@ -1,5 +1,88 @@
...
+  const dispatch: AppDispatch = useDispatch()
+  const navigate = useNavigate()
+
+  // Event Handlers
+  const formValidation = (): void => {
+    const validEmailAddressProvided = formData.email.match(VALID_EMAIL_REGEX)
+    const passwordFieldFilledOut = formData.password !== ''
+
+    if (validEmailAddressProvided && passwordFieldFilledOut) setButtonDisabled(false)
+    else setButtonDisabled(true)
```
yeezick, line 31: this is cool never seen an if...else like this

src/screens/SignIn/SignIn.tsx
```diff
@@ -1,5 +1,88 @@
...
+
+  const handleFormDataChange = (e: React.ChangeEvent<HTMLInputElement>): void => {
+    setFormData({ ...formData, [e.target.name]: e.target.value })
+  }
+
+  const handleSubmitForm = async (e: React.FormEvent): Promise<void> => {
+    e.preventDefault()
+
+    const response = await signIn(formData)
+    if (response?.message === "Invalid email or password" ?? false) return setInvalidCredentials(true)
```
yeezick, line 42: never seen ?? before so i looked into this too. accorting to [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) we're not using it properly as the left conditional would return either true or false, never null or undefined. it might be null or undefined if response comes back like that from the API which we'd have to clean up in an error handling epic. 

the conditional itself is also slightly hard to follow
`if ((response exists and has a message that equals 'invalid...' ) <= doesn't return 'null' or 'undefined' then return false) => return invalidCredentials(true)`

i'd suggest breaking this up a bit more to be easier to read. i trust you that it works and it was clever, but always prefer legibility over cleverness (unless youre a competitive programmer)

src/screens/SignIn/SignIn.tsx
```diff
@@ -1,5 +1,88 @@
...
+            <input
+              className={styles.input}
+              name="email"
+              id="email"
+              type="email"
+              onChange={handleFormDataChange}
+              required
+            />
+          </div>
+          <div className={styles.flex_column}>
```
yeezick, line 69: add a line break between input-divs, makes them easier to read

src/screens/SignIn/SignIn.tsx
```diff
@@ -1,5 +1,88 @@
...
+              id="password"
+              type="password"
+              onChange={handleFormDataChange}
+              required
+            />
+          </div>
+        </div>
+        <button disabled={buttonDisabled} type="submit">Go</button>
+      </form>
+      <p className={`${!invalidCredentials ? styles.hidden : styles.error_message}`}>{invalidCredentials && <span>Invalid Credentials</span>}</p>
```
yeezick, line 83: this is also very tricky. consider turning this into a component that receives props to decide on what to render.

and rather than set the styles to hidden, we should unmount/not return this component at all instead

src/utilities/api/apiConfig.js
```diff
@@ -23,3 +23,10 @@ api.interceptors.request.use(
     return Promise.reject(error);
   },
 );
+
```
yeezick, line 26: whitespace

yeezick: i'm curious what the interceptor below does too

src/index.css
```diff
@@ -1,3 +1,10 @@
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
```
yeezick, line 6: whitespace

yeezick (changes requested):
few more things

src/screens/SignIn/SignIn.module.css
zendrias: Yeah, I forgot I put one in the index.css.  I have always been a fan of CSS resets but it's but for debate. I'll remove the one I put in. Check Out [This Article](https://www.webfx.com/blog/web-design/should-you-reset-your-css/)

src/utilities/api/apiConfig.js
zendrias: Axios considers any API status response other than a 201 to be an error. So I'm just intercepting the error, and simply returning the error message.

yeezick:
I removed a console log and merged with develop, this is approved!

yeezick (approved):

PR #7, opened 2022-10-20
Feature/bc 327 display all projects

mendo94:
The projects are all displaying and the user is able to navigate to Browse Projects to view them all. Each project has a a link attached and once clicked, it navigates the user to another page and displays further details of that project on the screen. And when you look at the URL, you will see that the id renders from the params and displays the project’s id.


src/screens/Projects/ProjectDetails.tsx
```diff
@@ -0,0 +1,51 @@
...
+    project_owner: {
+      first_name: new String(),
+      last_name: new String(),
+      portfolio_link: new String(),
+    },
+  });
+  const params = useParams();
+
+  const fetchProject = async () => {
+    const displayOneProject = await getOneProject(params.id);
```
yeezick, line 25: move this into the useEffect since it's the only place it's getting called
it will get deleted with the callback vs. stay in this component scope

yeezick (dismissed):
Just resolve the one conversation, then you can self-merge

src/screens/Projects/ProjectDetails.tsx
mendo94: I did just that.

mendo94 (commented):
reviewed by Erick

yeezick (approved):

PR #8, opened 2022-10-23
Feature/bc 328 edit a project

aryannaf:
- populates fields based on selected project to be edited
- change information and update project
Functionality complete, pending final tests after schema to be updated with IDs so projects can be accessed


src/App.tsx
```diff
@@ -18,6 +19,7 @@ function App() {
           <Route path="/sign-up" element={<SignUp />} />
           <Route path="/sign-in" element={<SignIn />} />
           <Route path="/projects/create" element={<CreateProject />} />
+          <Route path="/projects/:id" element={<EditProject />} />
```
yeezick, line 22: use this route for the project's detail page, use `/projects/:id/edit` for their edit screen

src/layout/Nav/Nav.tsx
```diff
@@ -18,6 +18,9 @@ const Nav: React.FC = () => {
         <div>
           <Link to="/projects/create">Create Project</Link>
         </div>
+        <div>
+          <Link to="/projects/:id">Edit Project</Link>
```
yeezick, line 22: this link needs a hardcoded id to work, the edit form renders but the alert says there's no project found & the url ends in `/:id`

to get a generic id, you can call `getAllProjects` then use the id for the first project in that array

src/screens/EditProject/EditProject.scss
```diff
@@ -0,0 +1,53 @@
...
+
+  .checkbox {
+    width: 15px;
+  }
+
+  button {
+    margin: 15px;
+    text-transform: uppercase;
+  }
+}
```
yeezick, line 53: loving the alphabetized css

src/screens/EditProject/EditProject.tsx
```diff
@@ -0,0 +1,194 @@
...
+      status: string,
+      category: string,
+      title: string,
+      description: string,
+      skills: [],
+      desired_headcount: number,
+    ];
+    duration: string;
+    onChange?: (e: React.ChangeEvent<HTMLSelectElement>) => void;
+  }
```
yeezick, line 27: - curious about the type for roles.. does the object-like types within the `[ ]` define the types for each element within the "roles" array? 
- Koffi's merge into development has a project interface in `/src/utilities/interfaces/` that you replace with this type. look into [rina's PR](https://github.com/yeezick/bootcampr/pull/7/files#diff-7d32dddd29025e3430b288c438b146868131585dc989e47705ec6c220c8d965b) and see if you can merge your types together. the type for a "role" may have to become its own interface as well

src/screens/EditProject/EditProject.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  interface Props {
+    id?: EditProject;
+    project?: EditProject;
+  }
+
+
+const EditProject: React.FC<Props> = ({ id, project }) => {
+
+    const [fileSelected, setFileSelected] = useState<File>();
```
yeezick, line 37: actually super confused as to how the form values are being updated when the handler updates `setupdatedProject` but not the input values are set to `currentProject`

either way i think we can consolidate these states and set their types to the `EditProject` interface

yeezick (changes requested):
minor suggestions, i also see that the edit project form is very similar to [this pr](https://github.com/yeezick/bootcampr/pull/4/files#diff-f0e7b21f6031669de938397b19803a3ed7b6dc5f44ef0785e8dfadc7a180f835) so if any of those conversations on the `CreateProject` screen can be applied here please do so

src/App.tsx
mendo94: changed path on app.tsx

src/layout/Nav/Nav.tsx
mendo94: no longer necessary because redux store is now set up

src/screens/EditProject/EditProject.tsx
mendo94: no longer setting props here and using the project interface instead

src/screens/EditProject/EditProject.scss
mendo94: thank you

src/screens/EditProject/EditProject.tsx
mendo94: values are set to current project now

src/screens/CreateProject/EditProject.tsx
```diff
@@ -0,0 +1,169 @@
...
+        firstName: projectOwner.firstName,
+        lastName: projectOwner.lastName,
+        _id: projectOwner._id,
+      };
+      updateProjectForm(singleProject);
+    };
+    fetchProject();
+  }, []);
+
+  const handleInputChange = (e: React.ChangeEvent<HTMLInputElement | HTMLSelectElement | HTMLTextAreaElement>) => {
```
HectorIlarraza, line 40: 40-68 learned a lot from this and its very cool, nothing to say just admiration

HectorIlarraza (dismissed):
Perfect in my eyes Rina, I see no need to alter anything on ur PR

src/screens/CreateProject/EditProject.tsx
mendo94: wow thanks for your review!

yeezick (dismissed):
Excellent work @mendo94, well-written and concise. Great reviews @HectorIlarraza! That's the kind of eye it takes to maintain quality!

yeezick (approved):

PR #9, opened 2022-10-29
Feature/bc 324 sign up functionality

jasonmar08:
- Sign Up form with full functionality
- Sign Up Auth and API call with async thunk middleware managed in Redux store with extraReducers for different case scenarios (pending, fulfilled, rejected)
- Added a "Loading... ..." alert that shows up during a pending case. Can be replaced later on by a rotating circle or other icon, or removed if not preferred
- Created “useAppDispatch” and “useAppSelector” hooks in Redux store to allow global access to useDispatch and useSelector using Typescript

- Note: User is currently navigated to `/sign-in` after submitting Sign Up form. The navigation will need to be updated to `/users/id/account-setup` once Register User Info screen is created


src/screens/SignUp/SignUp.css
```diff
@@ -0,0 +1,57 @@
+.signup-container {
```
yeezick, line 1: good job on the ui! it came out very clean

however, we will be using sass (scss) over vanilla css. makes it slightly easier to work with without much overhead. 

please convert this file to sass & alphabetize the properties, practice adapted from [airbnb style guide](https://github.com/airbnb/css)

src/screens/SignUp/SignUp.css
```diff
@@ -0,0 +1,57 @@
...
+top: 33%;
+right: 44.5%;
+/* padding: 5px; */
+cursor: pointer;
+filter: invert(50%)
+}
+
+#pwd-reveal-gray:hover {
+filter: none;
+/* background-color: rgb(234, 234, 234);
```
yeezick, line 41: remove dead code

src/screens/SignUp/SignUp.tsx
```diff
@@ -1,7 +1,100 @@
...
-  return <div>SignUp</div>;
+  const navigate = useNavigate()
+  const dispatch = useAppDispatch()
+
+  const { isLoading, isSuccess } = useAppSelector((state) => state.auth)
+
+  const [inputType, setInputType] = useState('password')
+  const [pwdRevealIcon, setPwdRevealIcon] = useState('https://i.postimg.cc/zGQTSGmF/pngwing-com-1.png')
+  const [formValues, setFormValues ] = useState<SignUpInterface>({
+    email: "",
```
yeezick, line 19: objects would also be handy to alphabetize!

src/screens/SignUp/SignUp.tsx
```diff
@@ -1,7 +1,100 @@
...
 
 const SignUp: React.FC = () => {
-  return <div>SignUp</div>;
+  const navigate = useNavigate()
+  const dispatch = useAppDispatch()
+
+  const { isLoading, isSuccess } = useAppSelector((state) => state.auth)
+
+  const [inputType, setInputType] = useState('password')
+  const [pwdRevealIcon, setPwdRevealIcon] = useState('https://i.postimg.cc/zGQTSGmF/pngwing-com-1.png')
```
yeezick, line 17: feel free to use one of the [react-icons](https://react-icons.github.io/react-icons/search?q=eye) we have installed

src/screens/SignUp/SignUp.tsx
```diff
@@ -1,7 +1,100 @@
...
+    if (isSuccess) {
+      dispatch(reset())
+      setFormValues({
+        email: "",
+        first_name: "",
+        last_name: "",
+        password: "",
+        confirmPassword: ""
+      })
+      navigate('/sign-in')
```
yeezick, line 36: let's create a placeholder screen for the additional info page and route them to it here, so when that screen gets merged it would be instantly integrated

src/screens/SignUp/SignUp.tsx
```diff
@@ -1,7 +1,100 @@
...
+    }
+  }, [isSuccess, dispatch])
+  
+  const handleChange= (e: any) =>  {
+      setFormValues({ ...formValues, [e.target.name]: e.target.value })
+  }
+
+  const handleSubmit = async (e: any) => {
+    e.preventDefault()
+
```
yeezick, line 46: same here for dead code and whitespace

src/screens/SignUp/SignUp.tsx
```diff
@@ -1,7 +1,100 @@
...
+        first_name: "",
+        last_name: "",
+        password: "",
+        confirmPassword: ""
+      })
+      navigate('/sign-in')
+    }
+  }, [isSuccess, dispatch])
+  
+  const handleChange= (e: any) =>  {
```
yeezick, line 40: find the types of these form handlers by hovering over the inputs. vscode should give a dropdown with the types

src/screens/SignUp/SignUp.tsx
```diff
@@ -1,7 +1,100 @@
...
+  const handleSubmit = async (e: any) => {
+    e.preventDefault()
+    
+    // const newUser: SignUpInterface = (formValues)
+
+    dispatch(register(formValues))
+    // await signUp(formValues)
+  }
+  
+  const isFormValid = () => {
```
yeezick, line 53: i'd like to suggest creating a for loop that iterates over the keys of the formValues looking for a falsey value. 
maybe something along the lines of 
```
Object.keys(formValues).find(( input ) => !formValues[input])
``` 
let me know what you think

src/screens/SignUp/SignUp.tsx
```diff
@@ -1,7 +1,100 @@
...
+    return (!formValues.email || !formValues.first_name || !formValues.last_name || !formValues.password || !formValues.confirmPassword || formValues.password !== formValues.confirmPassword)
+  }
+  
+  const doPasswordsMatch = () => {
```
yeezick, line 57: since this is returning jsx, capitalize the first word too
the conditions below could also be simplified by checking for the `password` and `confirmPassword` first, then proceeding to check whether they match or not. since it's also an "either-or" situation it can just be an else statement

we also want to avoid inline styling
if this is not the last time we see info messages like this we might be tempted into making it a reusable component!

src/screens/SignUp/SignUp.tsx
```diff
@@ -1,7 +1,100 @@
...
+      setInputType('password')
+      setPwdRevealIcon('https://i.postimg.cc/zGQTSGmF/pngwing-com-1.png')
+    }
+  }
+
+  if (isLoading)
+    return <h1 style={{color: 'orange'}}>Loading... ...</h1>
+  
+  return (
+  <>
```
yeezick, line 81: no need for the fragment since you have a div

src/screens/SignUp/SignUp.tsx
```diff
@@ -1,7 +1,100 @@
...
+        <input type="text" name="last_name" placeholder="Last Name" onChange={handleChange} value={formValues.last_name} autoComplete="off"/>
+        <input type="email" name="email" placeholder="Email"  onChange={handleChange} value={formValues.email} autoComplete="off" required />
+        <div>
+          <img onClick={passwordReveal} src={pwdRevealIcon} alt='pwd' id={inputType === 'password' ? 'pwd-reveal-gray' : 'pwd-reveal'} />
+          <input type={inputType} name="password" placeholder="Password" onChange={handleChange} value={formValues.password} autoComplete="off" />
+        </div>
+        <input type={inputType} name="confirmPassword" placeholder="Confirm Password" onChange={handleChange} value={formValues.confirmPassword}  autoComplete="off"/>
+        <button type="submit" disabled={isFormValid()}>Create Account</button>
+      </form>
+      {doPasswordsMatch()}
```
yeezick, line 95: i'm curious about this, what does it do at the end of a file? would we be able to place this maybe `onBlur` of the confirmPassword input instead

src/utilities/Interface/SignUpInterface.ts
```diff
@@ -0,0 +1,7 @@
+export interface SignUpInterface {
+  email: string | undefined,
```
yeezick, line 2: same undefined note

src/utilities/redux/slices/users/userSlice.tsx
```diff
@@ -1,9 +1,11 @@
 import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { SignUpInterface } from '../../../Interface/SignUpInterface';
```
yeezick, line 2: fie imported but not used

src/utilities/redux/slices/users/userSlice.tsx
```diff
@@ -1,9 +1,11 @@
...
 import { UserInterface } from '../../../Interface/UserInterface';
 import { RootState } from '../../store';
 
 export interface UserState {
   authUser: UserInterface | null;
+
```
yeezick, line 8: whitespace

src/utilities/redux/slices/users/authSlice.tsx
```diff
@@ -0,0 +1,70 @@
+import { createSlice, createAsyncThunk, PayloadAction } from "@reduxjs/toolkit";
```
yeezick, line 1: okay so i'm very interested in the way you set this file up. it looks like you're error handling here which i'm very excited to hear your thoughts on.

i think it may be the start of good validation and i'd like to expand on this file with the whole team if possible. 

the only problem with this is that there are 2 slices for the same user. i'd like to suggest combining them together so that we could have the user data at the path `ui.auth.user` and the auth status' in `ui.auth`. 

then it would be good add visibility for the user so that when something goes wrong a message comes up. for example, i forgot to start my backend and when i completed the form i was confused why my clicks werent doing anything.

yeezick (changes requested):
@jasonmar08 hey, i know there were a ton of conversations in this PR but do not take them to heart, this is a great pull request and i like how you've developed this feature. most of my notes are just verbal linting, your logic is very solid.

if you'd like to get on a 1:1 to discuss or clarify anything about the above, feel free to ping me

i'm still interested in bringing up your user slice to the team though. 

beside that, just update your branch with develop and let me know when the conversations are resolved. you can close them as you take care of them.

src/utilities/redux/slices/users/userSlice.tsx
jasonmar08: resolved after pulling updated userSlice

jasonmar08:
Updates:

- Refactored validateForm and DoPasswordsMatch SignUp functions
- Updated SignUp screen to be managed by userSlice (Would like this to be reviewed)
- Set up placeholder for Additional Info screen so user is navigated to this screen following Sign Up (There is an issue retrieving user's first and last name, which might be due to a mismatch with interface keys and backend)

src/screens/SignUp/SignUp.scss
```diff
@@ -0,0 +1,86 @@
...
+}
+
+.pwd-reveal-gray:hover {
+  filter: none;
+}
+
+.signup-container {
+  display: flex;
+  flex-direction: column;
+  text-align: center;
```
yeezick, line 63: alphabetize pls

src/screens/SignUp/SignUp.tsx
```diff
@@ -1,7 +1,99 @@
...
-};
+  const navigate = useNavigate()
+  const dispatch = useAppDispatch()
+
+  const { isLoading, isSuccess } = useAppSelector((state) => state.ui.status)
+  const { _id } = useAppSelector((state) => state.ui.auth.user)
+
+  const [inputType, setInputType] = useState('password')
+  const [formValues, setFormValues ] = useState<SignUpInterface>({
+    confirmPassword: '',
```
yeezick, line 20: create a file called `src/utilities/constants.js` and export this variable from there to avoid hardcoding the objects and make them easier to update in the future

src/screens/SignUp/SignUp.tsx
```diff
@@ -1,7 +1,99 @@
...
+        email: '',
+        firstName: '',
+        lastName: '',
+        password: ''
+      })
+      navigate(`/users/${_id}/account-setup`)
+    }
+  }, [isSuccess, dispatch])
+  
+  const handleChange= (e: React.ChangeEvent<HTMLInputElement>) =>  {
```
yeezick, line 41: .

src/screens/SignUp/SignUp.tsx
```diff
@@ -1,7 +1,99 @@
...
+  
+  const validateForm = () => {
+    for (const [key, value] of Object.entries(formValues)) {
+      if (!value || (formValues.password !== formValues.confirmPassword)) {
+        return true
+      }
+    }
+  }
+  
+  const DoPasswordsMatch = () => {
```
yeezick, line 58: this is not JSX so name should not be capital

src/screens/SignUp/SignUp.tsx
```diff
@@ -1,7 +1,99 @@
...
+    dispatch(register(formValues))
+  }
+  
+  const validateForm = () => {
+    for (const [key, value] of Object.entries(formValues)) {
+      if (!value || (formValues.password !== formValues.confirmPassword)) {
+        return true
+      }
+    }
+  }
```
yeezick, line 56: okay, so these are the use cases we need to handle:
- the passwords should have their own state and status for when they match or dont match
  - if they're empty we show nothing 
  - if they dont match, we show they dont match
  - if they do match, we show they match 

the button itself should be disabled if any of the fields in the validate form are empty and if the password fields dont match 

this function disturbed me because it handles 2 different things at once:
- check if form is empty 
- validate passwords

by separating those concerns into their own state and functions, the code is decoupled, more manageable, and easier to read.

src/screens/SignUp/SignUp.tsx
```diff
@@ -1,7 +1,99 @@
...
+  
+  const DoPasswordsMatch = () => {
+    if (formValues.password && formValues.confirmPassword) {
+      if (formValues.password !== formValues.confirmPassword) {
+        return <h4 className='pwd-mismatch'>Passwords do not match</h4>
+      } else {
+        return <h4 className='pwd-match'>Passwords match!</h4>
+      }
+    }
+  }
```
yeezick, line 66: this is fine for now, just remove the conditional chaining and destructure the pw and cpw so this function reads a bit easier:
```js
  const DoPasswordsMatch = () => {
    const {confirmPassword, password} = formValues
    if (!confirmPassword | !password} return null 
    
      if (password !== confirmPassword) {
        return <h4 className='pwd-mismatch'>Passwords do not match</h4>
      } else {
        return <h4 className='pwd-match'>Passwords match!</h4>
      }
  }
```

i would even suggest taking a look at [this article on multiple conditions](https://www.freecodecamp.org/news/learn-react-conditionals/#:~:text=statements%20for%20multiple-,conditions,-What%20if%20we), but both would be valid unless our conditions grew. a switch case would scale better, but this feature specifically may not ever have more than 3 conditions

you can also call this component like a regular `<Component />` below

yeezick:
changed this to merge into my branch for now since it's carrying my pr with it. if i merge to develop before this pr gets finished, we will repoint to develop. 

great work jason!

src/utilities/types/UserInterface.ts
```diff
@@ -28,12 +28,20 @@ export interface UiSliceInterface {
...
+      status?: boolean;
+      message?: string;
     };
   };
 }
 
+export interface SignUpInterface {
+  confirmPassword: string | undefined,
+  email: string | undefined,
+  firstName: string | undefined,
```
yeezick, line 40: if a property is optional use `?` instead. avoid null or undefined types

src/utilities/redux/slices/users/authSlice.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { SignUpInterface } from "../../../types/UserInterface";
+import { signUp } from "../../../api/users";
```
yeezick, line 3: this file can be deleted

PR #11, opened 2022-11-01
Feature/bc 260 user dashboard

Bestnogm1:
added the dashboard and created the dashboard folder and file added the routes to see the detail of the project


src/App.tsx
```diff
@@ -6,6 +6,7 @@ import { SignIn } from './screens/SignIn/SignIn';
 import { SignUp } from './screens/SignUp/SignUp';
 import { Nav } from './layout/Nav/Nav';
 import { Layout } from './layout/Layout';
+import Dashboard from './screens/Dashboard/Dashboard';
```
zendrias, line 9: Keep imports consistent with curly braces

Bestnogm1 (commented):
fixed import

src/utilities/types/UserInterface.ts
```diff
@@ -1,14 +1,14 @@
...
+export interface UserInterface {
   bio: string;
   declinedProjects?: ProjectInterface[];
   email: string;
   firstName: string;
   interestedProjects?: ProjectInterface[];
   lastName: string;
   linkedinUrl: string;
-  memberOfProjects?: ProjectInterface[];
+  memberOfProjects?: ProjectInterface[] | null;
```
yeezick, line 11: since `memberOfProjects` already has a `?` we don't need to add null. The api will just not return it.

src/utilities/Interface/ProjectInterface.ts
```diff
@@ -0,0 +1,15 @@
+export interface ProjectInterface {
```
yeezick, line 1: we already have a ProjectInterface in `types/` ?

src/screens/Dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,59 @@
...
+    fetchUsers();
+  }, []);
+
+  useEffect(() => {
+    setCurrentUser(allUser);
+  }, [allUser]);
+  console.log(user);
+
+  return (
+    <>
```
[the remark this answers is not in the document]
yeezick: Also, simplify this component by adding an if statement before the render instead of conditionally rendering the return 
if (!user) return <p> no project yet </p>

return (<rest of file since user exists/>)

yeezick (changes requested):
@zendrias please be more thorough in your reviews. 
compare with other files in the codebase to identify how to keep our style guides and best practices consistent. 
@Bestnogm1 good work, just needs some refinement & merge with dev.

PR #16, opened 2022-11-16
Hotfix/bc 351 no access to mdb

yeezick:
adds logic to default to localhost if BACKEND_ENV is not cloud

ideally, env vars over conditional like this


HectorIlarraza (approved):
Everything looks good, I see you tweaked it and cleaned it up on apiConfig file to adjust to future plans noice.

PR #26, opened 2022-12-11
Feature/bc 308 create roles

yeezick:
Completes [BC-308: create roles](https://bootcamper.atlassian.net/browse/BC-308?atlOrigin=eyJpIjoiN2Y5YWU1YmUzZDY0NGU3M2FiYTExMWUwYmMxY2YxNDMiLCJwIjoiaiJ9)
This ticket is meant to be tested and merged with its [backend counterpart](https://github.com/yeezick/bootcampr-backend/pull/29)

- restructures interfaces to support role types (engineering, design, etc) 
- CreateRole component handles the form for the new role 
- ViewRoles handles rendering the roles of the project form, which is updated after a role is created 
- Also creates some helper files for StringUtils, starting with capitalizing a title


src/screens/Roles/ViewRoles.tsx
```diff
@@ -0,0 +1,48 @@
...
+    <div>
+      {applicants.map(({ firstName, lastName }) => (
+        <p>{firstName + ' ' + lastName}</p>
+      ))}
+    </div>
+  );
+};
+
+// If you see `roles[type as keyof typeof roles]` => https://stackoverflow.com/questions/73714390/ts7053-element-implicitly-has-an-any-type-because-expression-of-type-string
+const MapRoles = ({ role }: MapRolesProps) => {
```
mendo94, line 16: Getting an 'each child should have a unique key prop' error for mapRoles

src/screens/Roles/ViewRoles.tsx
```diff
@@ -0,0 +1,48 @@
+import { generateTitle } from '../../utilities/helpers/stringUtils';
+import { InterestedApplicantsProps, MapRolesProps, ViewRolesProps } from './PropInterfaces';
+import './CreateRole.scss';
+
+const MapInterestedApplicants = ({ applicants }: InterestedApplicantsProps) => {
```
mendo94, line 5: Getting an 'each child should have a unique key prop' error for MapInterested applicants

src/screens/CreateProject/EditProject.tsx
```diff
@@ -8,14 +8,15 @@ import { emptyProject } from '../../utilities/data/projectConstants';
 import { getOneUser } from '../../utilities/api/users';
 import { getOneProject, editProject, deleteProject } from '../../utilities/api/projects';
 import './CreateProject.scss';
+import { CreateRole } from '../Roles/CreateRole';
```
mendo94, line 11: This component is giving 'a component is changing a controlled input to be uncontrolled' error

src/screens/CreateProject/EditProject.tsx
```diff
@@ -8,14 +8,15 @@ import { emptyProject } from '../../utilities/data/projectConstants';
 import { getOneUser } from '../../utilities/api/users';
```
mendo94, line 8: Are you getting the same error when the edit project screen loads? "Cast to ObjectId failed for value "undefined" (type string) at path "_id" for model "User"" because of the id being undefined, it will not let me complete the create role request, just crashes with error "Cannot convert undefined or null to object"

src/utilities/types/ProjectInterface.ts
```diff
@@ -6,11 +6,23 @@ export interface ProjectInterface {
   meeting_cadence: string;
   overview: string;
   project_owner: string | UserInterface;
-  roles?: string[];
+  roles: {
```
mendo94, line 9: Wow, hopefully this is the fix to my error - putting interface[] on each variable!!

src/screens/CreateProject/EditProject.tsx
mendo94: Screen Shot 2022-12-11 at 7 11 50 PM

mendo94 (changes requested):
Wow your Typescript abilities have grown immensly. Can;t wait to merge your branch and get some breakthroughs over here!! Also, getting a few errors and was wondering if you were getting them on your end or not when rendering the edit project screen.

src/screens/CreateProject/EditProject.tsx
yeezick: This is an issue with `<EditProject />`'s file input, after commenting it out the warning went away. 
It's because that input doesn't have a value but its value changes with state; the recommended approach is to [set this input's value via useRef](https://reactjs.org/docs/uncontrolled-components.html#the-file-input-tag)

Instead, we should open a bug ticket or add it to the AC of Koffi's profile image PR

yeezick:
This PR is considered outdated based on the new direction of the app. 
This PR has not resolved all of its conflicts and is not yet fully functioning. 
Only being merged to close the PR and preserve the work in case it's ever needed again.

PR #27, opened 2022-12-17
Feature/bc 260 user dashboard 2.0

Bestnogm1:
So I scratched the old user dashboard  and added a new 2.0 version of the dashboard was also able to meet every criteria of the user dashboard ticket Feature/bc 260 user dashboard 2.0


mendo94 (commented):
This PR meets the acceptance criteria on the jira ticket, which is doing most of the same things as "My Profile" screen besides having the "Edit my profile" button, so requesting @yeezick look at this ticket to see if there is a need for "My Profile" screen then.

yeezick:
I feel like I remember having a discussion on what the need for the user dashboard... I don't remember what came of it though. If I'm mistaken then I agree with @mendo94, we might not have a use case for it until UX lets us know what to render.
Does anybody remember that discussion? Or am I mistaken?

yeezick:
@mendo94  makes a good point. Closing this PR until dashboard is revisited with UX team 
Thank you for working on it @Bestnogm1

PR #28, opened 2022-12-18
Bc 348 custom profile links

mendo94:
After speaking with Erick and the design team, we agreed that this ticket is no longer needed. We will add one more attribute to the user model, which will be a github URL if they are a software engineer. After speaking with design, they are content with the url links (linkedin, portfolio link) that are already on the user model.

Updating the user's profile will work, but there is an error coming from the latest changes to the backend, "error: 
"updatingImage is not defined" because the logic is not on the front end yet.


src/screens/UserProfile/EditProfile.tsx
```diff
@@ -14,13 +14,30 @@ export const EditProfile: React.FC = () => {
...
+    lastName,
+    linkedinUrl,
+    githubUrl,
+    portfolioUrl,
+    profilePicture,
+    role,
+    _id: userId,
+  } = userForm;
+
+  console.log(authUser);
```
yeezick, line 29: console log

yeezick (dismissed):
a scoundrel got away from u, besides that...


ping me to reapprove when ready

src/screens/UserProfile/EditProfile.tsx
mendo94: removed the culprit

yeezick (approved):

PR #29, opened 2022-12-18
Feature/bc 260 profile images

Bestnogm1:
Added feature to be able to add profile images and and edit it as well


src/screens/PreviewsUserImage/PreviewsUserImage.tsx
```diff
@@ -0,0 +1,15 @@
+export const PreviewsUserImage = ({ previewImage, authUser }: any) => {
```
mendo94, line 1: can any be changed to a specific type?

src/utilities/api/users.ts
```diff
@@ -14,15 +16,20 @@ export const getAllUsers = async () => {
 export const getOneUser = async (id: any) => {
   try {
     const res = await api.get(`/users/${id}`);
+    console.log(res.data);
```
mendo94, line 19: good to remove

mendo94 (changes requested):
minor fixes, amazing work as always!!

src/screens/RegisterUserInfo/RegisterUserInfo.tsx
```diff
@@ -37,8 +43,11 @@ export const RegisterUserInfo: React.FC = () => {
     setUserForm({ ...userForm, [name]: value });
   };
 
-  const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+  const handleSubmit = async (e: React.FormEvent<HTMLFormElement>) => {
     e.preventDefault();
+
```
yeezick, line 48: since it's short we can get rid of the whitespace around the `if`. we want to think of code like a book, sometimes it makes sense to have short paragraphs (separating bits of logic), but when the whole block is 1 short paragraph, we can just let it read

src/utilities/api/users.ts
```diff
@@ -14,15 +16,20 @@ export const getAllUsers = async () => {
...
     return res.data;
   } catch (error) {
     throw error;
   }
 };
 
-export const updateUser = async (id: any, userUpdate: any) => {
+export const updateUser = async (
+  id: string | undefined,
```
yeezick, line 27: string shouldn't be undefined, in this function, if it didn't have the id it would fail
haven't seen ur backend pr but nice `imageWasUpdated` key to the req.body `B^)`  nice n orthogonal

yeezick:
nice pr

src/utilities/api/users.ts
mendo94: the backend PR for this ticket was already merged into dev two weeks ago I believe

PR #35, opened 2023-02-02
Feature/clean up applicant code

yeezick:
Removing all screens and components related to projects or applications. Everything related to projects should be from scratch.


gooliajulia:
Should we remove the "Browse Projects" and "Create Project" links from the nav bar too in this PR? They're broken now anyway after these changes

yeezick:
good catch, thank you

gooliajulia (approved):

PR #39, opened 2023-02-22
Feature/bc 260 profile images2.0

Bestnogm1:


src/layout/Nav/Nav.tsx
```diff
@@ -81,7 +81,14 @@ export const Nav = () => {
               <span>{notificationCount}</span>
             </div>
           )}
-          <div className='image'></div>
+          <div className='image'>
```
yeezick, line 84: if there's no reason to wrap this `` in a div then we should remove it and apply the styles to it via classes

src/layout/Sidebar/Sidebar.tsx
```diff
@@ -11,7 +11,7 @@ import './Sidebar.scss'
 
 export const Sidebar = () => {
   const authUser = useAppSelector(selectAuthUser)
-  const { _id: userId, firstName, lastName } = authUser
+  const { _id: userId, firstName, lastName, profilePicture } = authUser
```
yeezick, line 14: just realized this line is redundant & can be simplified, would you mind adding this for me instead?  
`const { _id: userId, firstName, lastName, profilePicture }  = useAppSelector(selectAuthUser)`

src/layout/Sidebar/Sidebar.tsx
```diff
@@ -36,9 +36,16 @@ export const Sidebar = () => {
...
       <div className='current-user'>
-        <div className='image'></div>
+        <div className='image'>
+          <img
+            src={profilePicture}
+            alt={`${firstName} profile picture`}
+            width='110'
+            height='110'
+          />
+        </div>
```
yeezick, line 48: remove dead code & same comment for the image tag

src/layout/Nav/Nav.tsx
yeezick: also, use the user's full name with a `'` at the end

src/screens/Auth/RegisterUserInfo/RegisterUserInfo.tsx
```diff
@@ -8,11 +8,16 @@ import {
 import { UserInterface } from 'utilities/types'
 import { emptyUser } from 'utilities/data/userConstants'
 import './RegisterUserInfo.scss'
+import PreviewUserImage from '../PreviewUserImage/PreviewUserImage'
```
yeezick, line 11: no default exports, they lead to poor discoverability for the compilers/interpreters as well as for tools like vscode.
for example in large codebases when i try to import, vscode doesn't make a suggestion to autocomplete the path bcs it can't find it

src/screens/Auth/RegisterUserInfo/RegisterUserInfo.tsx
```diff
@@ -43,9 +48,15 @@ export const RegisterUserInfo: React.FC = () => {
...
+  const handleSubmit = async (e: React.FormEvent<HTMLFormElement>) => {
     e.preventDefault()
-    dispatch(updateProfile(userForm))
+    let imageWasUpdated = false
+    if (profileImageFile) {
+      await createUserImage(profileImageFile, authUser._id)
+      imageWasUpdated = true
+    }
+    dispatch(updateProfile({ ...userForm, imageWasUpdated: imageWasUpdated }))
+    console.log({ ...userForm, imageWasUpdated: imageWasUpdated })
```
yeezick, line 59: console log

src/screens/Auth/RegisterUserInfo/RegisterUserInfo.tsx
```diff
@@ -54,23 +65,16 @@ export const RegisterUserInfo: React.FC = () => {
...
+            authUser={{ profilePicture: profilePicture }}
+          />
           <label>Profile Photo:</label>
-          <input
-            onChange={handleChange}
-            type='text'
-            name='profilePicture'
-            placeholder='Profile Photo'
-            value={profilePicture}
+          <RegisterUserImage
```
yeezick, line 73: can we rename this to `<UploadUserImage/>`

src/screens/Auth/SignUp/SignUp.tsx
```diff
@@ -108,7 +115,14 @@ export const SignUp: React.FC = () => {
...
       : setConfirmPasswordInputType('password')
   }
+  const randomHexCodeColor = () => {
+    const hexCodes = ['442288', '6CA2EA', 'B5D33D', 'EB7D5B', 'FED23F']
+    return hexCodes[Math.floor(Math.random() * 5)]
+  }
 
+  function createProfileImage(name: string, lastName: string): string {
+    return `https://ui-avatars.com/api/?name=${name}+${lastName}&background=${randomHexCodeColor()}&color=FFFFFF&font-size=0.49`
+  }
```
yeezick, line 125: hell yeah on this service

src/utilities/api/users.ts
```diff
@@ -25,7 +26,11 @@ export const getOneUser = async (id: any) => {
   }
 }
 
-export const updateUser = async (id: any, userUpdate: any) => {
+export const updateUser = async (
+  id: any,
+  userUpdate: any,
+  imageWasUpdated: boolean | null
+) => {
```
yeezick, line 33: we're passing in `imageWasUpdated` which doesn't seem to be used here? 
would u mind typing the other params too please, i know it was previously set to any

yeezick (changes requested):
Brilliant PR Koffi, very well done
most of my comments are about linting/semantics; the logic looks solid to me 
one final suggestion:
the components `<PreviewUserImage />` and `<RegisterUserImage />` are just components and not screens. could you move them into the components directory `components/UserImage/`

PR #45, opened 2023-04-18
Feature/BC-464 Basic UI

gooliajulia:
* enable user to select the timezone
* render row for every week day
* on selection of a weekday, render time slot inputs (default 9-5)
* enable user to customize time slots (just single)
* include icons (no functionality)


src/components/Availability/subcomponents/DayAvailabilityInputBanner.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  return (
+    <div>
+      <div className='day-availability-input-banner'>
+        <div className='left-banner'>
+          <div className='check-day'>
+            <Checkbox
+              name={day}
+              onChange={e => handleCheck(e)}
+              sx={{ color: '#022888', '&.Mui-checked': { color: '#022888' } }}
```
yeezick, line 31: not asking for a change yet but would be nice to standardize how we style mui components & see if we can consolidate our approach to a single one 
inclined to stay away from `sx` because it's the same as inline styling
even if we needed to do it it'd be preferable to create variables for their objects over polluting the component's props
https://mui.com/material-ui/customization/how-to-customize/

src/components/Availability/subcomponents/DayAvailabilityInputBanner.tsx
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  return (
+    <div>
+      <div className='day-availability-input-banner'>
+        <div className='left-banner'>
+          <div className='check-day'>
+            <Checkbox
+              name={day}
+              onChange={e => handleCheck(e)}
```
yeezick, line 30: no need for anon function unless you're passing additional params to the function `onChange={handleCheck}`

src/components/Availability/subcomponents/SelectTimeInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+  '& .MuiSvgIcon-root': { display: 'none' },
+  backgroundColor: '#fefefe',
+  '.MuiOutlinedInput-notchedOutline': { border: 0 },
+  width: '87px',
+  fieldset: {
+    border: 'none !important',
+    outline: 'none !important',
+  },
+  elevation: '0',
+}
```
yeezick, line 57: nice, alphabetize/organize though

src/components/Availability/subcomponents/SelectTimeInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+    })
+  }
+
+  return (
+    <Select
+      defaultValue={slot[index]}
+      inputProps={{ sx: { padding: '8px 13px !important' } }}
+      MenuProps={menuPropsSX}
+      name={`${day}-${idx}-${index}`}
+      onChange={e => handleTimeChange(e)}
```
yeezick, line 31: same ab anon function here

yeezick (changes requested):
stellar work , legible and great to follow
love the implementation for this feature
minor comments before approval

src/components/Availability/subcomponents/SelectTimeInput.tsx
gooliajulia: this function is utilizing the event (e) prop to get current input value. lmk if you'd approach another way!

yeezick: react events always infer the event object will be the first prop of the event's function 
the onChange would run the same without the anon function

src/components/Availability/subcomponents/DayAvailabilityInputBanner.tsx
gooliajulia: cool yeah I wasn't sure best way to handle these. just pushed a change converting them all to variables for now

src/components/Availability/subcomponents/SelectTimeInput.tsx
gooliajulia: oooooohhhh shit!! fixed

yeezick (approved):

PR #54, opened 2023-06-05
Feature/bc 475 view and edit profile component

HectorIlarraza:


src/interfaces/UserInterface.ts
```diff
@@ -6,13 +6,15 @@ export interface UserInterface {
   declinedProjects?: ProjectInterface[]
   email: string
   firstName: string
-  githubUrl: string
   interestedProjects?: ProjectInterface[]
   lastName: string
-  linkedinUrl: string
+  links: {
+    githubUrl: string
```
yeezick, line 12: this should be optional as a UXD won't have this url

src/screens/Auth/RegisterUserInfo/RegisterUserInfo.tsx
```diff
@@ -36,7 +34,18 @@ export const RegisterUserInfo: React.FC = () => {
...
+      setUserForm({
+        ...userForm,
+        links: {
+          ...userForm.links,
+          [name]: value,
+        },
+      })
+    } else {
+      setUserForm({ ...userForm, [name]: value })
+    }
```
yeezick, line 48: confused about this statement as we're checking a static array if it includes the name of the input then assigning it userForm.links 

while in the else statement we're skipping `.links` and setting the input name as `userForm[name]: value`

src/screens/UserProfile/EditProfile.scss
```diff
@@ -1,23 +1,130 @@
...
+
+  &__container, &__form, &__label {
+    @include flex(column, inherit, inherit);
+  }
+
+  &__backContainer {
+    @include flex();
+    margin: 17px 0 0 24px;
+
+    p {
```
yeezick, line 29: since all other classes branch off `&__` let's move the plain selectors above the class extensions

src/screens/UserProfile/EditProfile.tsx
```diff
@@ -35,11 +38,30 @@ export const EditProfile: React.FC = () => {
...
+      updateUserForm({
+        ...userForm,
+        links: {
+          ...userForm.links,
+          [name]: value,
+        },
+      })
+    } else {
+      updateUserForm({ ...userForm, [name]: value })
+    }
```
yeezick, line 60: same comment as above

src/screens/UserProfile/UserProfile.tsx
```diff
@@ -16,32 +22,157 @@ export const UserProfile = () => {
     navigate(`/users/${authUser._id}/edit`)
   }
 
+  console.log(authUser)
```
yeezick, line 25: remove log

src/screens/UserProfile/UserProfile.tsx
yeezick, line 1: let's remove the conditionals that have dummy info
only a logged in user should be able to reach this page

src/screens/UserProfile/UserProfile.tsx
```diff
@@ -16,32 +22,157 @@ export const UserProfile = () => {
...
+
+          <div className='userProfile__linkItem'>
+            <RiGithubLine className='userProfile__icons' />
+            <div className='userProfile__link'>
+              <h3>Github</h3>
+              {(authUser.role === 'Software Engineer' ||
+                authUser.role === 'UX Designer') &&
+              authUser.links &&
+              authUser.links.githubUrl ? (
+                <a
```
yeezick, line 125: role should only check if the user is an SWE

src/utils/data/userConstants.ts
```diff
@@ -43,10 +43,12 @@ export const initialState: UiSliceInterface = {
       bio: '',
```
yeezick, line 43: combine initial state and empty user so that initialState.auth.user = emptyUser

src/utils/redux/slices/userSlice.ts
```diff
@@ -64,6 +66,12 @@ export const updateProfile = createAsyncThunk(
     try {
       const res = await updateUser(user._id, user)
       if (res) {
+        res.links = {
+          portfolioUrl: user.links.portfolioUrl,
+          githubUrl: user.links.githubUrl,
+          linkedinUrl: user.links.linkedinUrl,
+        }
```
yeezick, line 73: we shouldn't have to hardcode this if the user is submitting this info themselves from their form

yeezick (changes requested):
good pr! questions/comments above

src/screens/UserProfile/UserProfile.tsx
```diff
@@ -16,32 +22,157 @@ export const UserProfile = () => {
...
+            <IconButton
+              aria-label='change profile pic'
+              className='userProfile__cameraIcon'
+            >
+              <CameraAltOutlinedIcon className='userProfile__imageChange' />
+            </IconButton>
+          </div>
+          <div className='userProfile__title'>
+            <h2>
+              {authUser.firstName ? authUser.firstName : 'Jeanine'}{' '}
```
yeezick, line 52: `authUser.firstName && <p> {authUser.firstName} </p>

src/screens/Auth/RegisterUserInfo/RegisterUserInfo.tsx
HectorIlarraza: I now know why i did it this way its because 

```
const {
    bio,
    firstName,
    lastName,
    links: { githubUrl, linkedinUrl, portfolioUrl },
    profilePicture,
    role,
    _id: userId,
  } = userForm```

has a nested object and in order to properly access it from the userForm i would need to look in it via the array. If it is, it means the changed field is one of the properties inside links, and so the value of this property is updated inside userForm.links.

if the name of the input field isn't found in the array, it means that the changed field is a top-level property in userForm (like firstName, lastName, or bio). which in terms updates the top-level property directly in userForm, without going through links.

Because userForm includes nested objects, the function needs to differentiate between changes to top-level properties versus nested properties. By doing so, it ensures that all properties, whether top-level or nested, are accurately tracked and updated in response to user input.

I did update the function to make it easier and cleaner to read so let me know if you like the updated one better or if u don't like this approach. We can flatten the object but that would require changing the interface and backend as well.

Comment (yeezick) [reply src/screens/Auth/RegisterUserInfo/RegisterUserInfo.tsx]: i see, thanks for explaining it to me!
rather than hardcoding a array, let's try 
```
    if (Object.keys(userform.links).includes(name)) {
      setUserForm({
        ...userForm,
        links: {
          ...userForm.links,
          [name]: value,
        },
      })
    }
``` 
so that it's dynamic to whatever links the user adds and only checks the links they do have

File: src/screens/Auth/RegisterUserInfo/RegisterUserInfo.tsx [hunk 10]
```diff
@@ -30,20 +30,24 @@ export const RegisterUserInfo: React.FC = () => {
     }
   }, [authUser])
 
+  const nestedProperties = Object.keys(userForm.links)
```
yeezick, line 33: disregard other comment as u already handled it here! 
inly suggestion is to rename var to nestedlinks as its more semantic

src/screens/Auth/RegisterUserInfo/RegisterUserInfo.tsx
yeezick: sounds good to me! thanks for explaining

yeezick (dismissed):

yeezick (approved):

PR #62, opened 2023-06-28
Feature/bc 470 enable chat socket

jasonmar08:
- Incorporated socket.io for real-time messaging, but may need to be refactored since multiple api calls are being made with this iteration.


yeezick (approved):
LGTM!

PR #67, opened 2023-07-20
created a reusable snackbar class to be able to be used any where in …

Bestnogm1:
created the snack bar to be used as a reusable component `  isOpen?: boolean
  message?: string
  duration?: number
  vertical?: 'top' | 'bottom'   // default: bottom
  horizontal?: 'left' | 'center' | 'right'  // default: left
  snackbarStyle?: string 
}`  these are some of the properties that someone needs to input when passing in as props  to the <SnackBarToastProps/> component

### DIFF
--- a/src/components/Kanban/AllTickets.tsx
+++ b/src/components/Kanban/AllTickets.tsx
@@ -151,6 +151,8 @@ export const AllTickets = ({ projectTracker }) => {
                               {...provided.dragHandleProps}
                             >
                               <TicketDetail
+                                openSnackBar={openSnackBar}
+                                setOpenSnackBar={setOpenSnackBar}
                                 ticketDetail={ticketDetail}
                                 getAllTicket={getAllTicket}
                                 setGetAllTicket={setGetAllTicket}
--- a/src/components/Kanban/AllTickets.tsx
+++ b/src/components/Kanban/AllTickets.tsx
@@ -17,11 +17,6 @@ import { SnackBarToastInterface } from 'interfaces/SnackBarToast'
 
 export const AllTickets = ({ projectTracker }) => {
   const { id } = useParams()
-  const [openSnackBar, setOpenSnackBar] = useState<SnackBarToastInterface>({
-    isOpen: false,
-    message: '',
-    duration: 0,
-  })
   const [getAllTicket, setGetAllTicket] = useState(
     projectTracker?.projectTracker
   )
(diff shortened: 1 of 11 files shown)


src/components/SnackBarToast/SnackBarToast.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react'
+import Snackbar from '@mui/material/Snackbar'
+import { SnackBarToastInterface } from 'interfaces/SnackBarToast'
+import { Alert } from '@mui/material'
+import CheckIcon from '@mui/icons-material/Check'
+import ErrorOutlineIcon from '@mui/icons-material/ErrorOutline'
+import WarningAmberIcon from '@mui/icons-material/WarningAmber'
```
yeezick, line 7: consolidate imports with 
`import {checkIcon, ErrorOutlineIcon, WarningAmberIcon} from '@mui/icons-material'`

src/components/SnackBarToast/SnackBarToast.tsx
```diff
@@ -0,0 +1,58 @@
...
+export interface SnackBarToastProps {
+  snackbarDescription?: React.Dispatch<
+    React.SetStateAction<SnackBarToastInterface>
+  >
+  snackBar?: SnackBarToastInterface
+  snackbarStyle?: string
+}
```
yeezick, line 14: move interface to interfaces directory

src/components/Kanban/AllTickets.tsx
```diff
@@ -12,10 +12,16 @@ import {
...
+import { SnackBarToastInterface } from 'interfaces/SnackBarToast'
 
 export const AllTickets = ({ projectTracker }) => {
   const { id } = useParams()
-
+  const [openSnackBar, setOpenSnackBar] = useState<SnackBarToastInterface>({
+    isOpen: false,
+    message: '',
+    duration: 0,
+  })
```
yeezick, line 24: the ability to open snack bars and populate with information should be available from any place in the app so it's probably best to set this up in redux so that we can call on the snackbar anywhere

yeezick (changes requested):
great work! you made the implementation look super simple

before we merge we should move the ability to open & populate snackbars to redux so that we can use the same component for other actions like successfully creating calendar events or error toasts when an api call fails 

i know it wasn't in the scope of the ticket but could you add a button to the sidemenu that onClick will trigger a snackbar with any message you want (like a joke or inspo message)?

src/components/SnackBarToast/SnackBarToast.tsx
Bestnogm1: tried this and its just not working

yeezick (approved):

PR #72, opened 2023-08-09
Feature/bc 527 email user unread messages alert

jasonmar08:


src/components/ChatDialog/Conversations/Conversations.tsx
```diff
@@ -213,14 +210,14 @@ const ConversationThumbnail = ({
         <div className='thumbnail-right'>
           <div className='thread-details-grid'>
             <h5
-              className={lastMessage && !isMessageRead ? 'unread-message' : ''}
+              className={lastMessage && messageIsUnread ? 'unread-message' : ''}
```
mendo94, line 213: I see this is repeated a couple of times. For better readability, can this be moved into a helper function?

mendo94 (approved):
As always great work! Just added one suggestion here but it is up to you.

mendo94 (approved):

PR #73, opened 2023-08-14
Bc 315 navbar

mendo94:
As a user, I want to be able to navigate to the following pages:

landing/home page (bc icon)

project portal

how bc works

about us

As well as being able to:

open the chat dialog 

open the account dropdown

From anywhere on the site so that I can easily access these features and pages. 

AC’s
Update nav links to reflect those mentioned above & still in figma 

Design nav to be like the figma 

Screen Shot 2023-08-13 at 10 22 49 PM


src/layout/Nav/Nav.tsx
```diff
@@ -122,44 +116,45 @@ const AuthorizedNavLinks = ({
...
-        {(visibleChat || !visibleChat) && (
-          <ChatIconBadge
-            isChatBadgeUpdated={isChatBadgeUpdated}
-            setIsChatBadgeUpdated={setIsChatBadgeUpdated}
+      <div className='message-container'>
+        <div className='messages-icon' ref={chatRef}>
+          <BsFillChatLeftTextFill
+            size={23}
+            className='chat-icon'
+            onClick={() => toggleChatBox()}
```
yeezick, line 124: suggestion to turn this into a [named function vs anon callback ](https://www.geeksforgeeks.org/difference-between-anonymous-and-named-functions-in-javascript/#)

src/layout/Nav/Nav.scss
```diff
@@ -99,38 +108,35 @@ nav {
...
+    padding: 6px 12px;
+    text-decoration: none;
   }
 
   .log-in {
-    background-color: #fefefe;
-    border-radius: 10px;
-    color: #000;
-    padding: 15px 30px;
+    background: var(--white-ffffff, #fefefe);
```
yeezick, line 131: wondering about these vars, why do we have to declare the hex code and a var name?

src/App.tsx
```diff
@@ -15,13 +15,18 @@ import { AllProjects, CreateProject, ProjectDetails } from 'screens/Project'
 import { AvailabilityDemoScreen } from 'screens/TempFeatures/Availability'
 import { CalendarScreen } from 'screens/Calendar/Calendar'
 import './App.css'
+import AboutUs from 'layout/Nav/Links/AboutUs'
+import HowTo from 'layout/Nav/Links/HowTo'
+import ProjectPortal from 'layout/Nav/Links/ProjectPortal'
```
yeezick, line 20: We should create the AboutUs and HowTo screens in the screens/landing directory

Project portal is unique and not an actual page but more of a portal to the project features when a user is logged in

For example, there is no sidebar in the landing, about, or how to screens but when the user clicks "project portal" they are routed to the project details screen with the sidebar always open.

yeezick: and they should be exported as consts, not default exports

yeezick (changes requested):
only one worth waiting for is the comment about screens and project portal

src/App.tsx
mendo94: Okay agree about the project portal, the screens I am not following the example you provided. Send help

src/layout/Nav/Nav.tsx
mendo94: moved to a named function

src/layout/Nav/Nav.scss
mendo94: you never know if you need a backup to your backup color, jk I copied the style straight from Figma. Removed vars.

src/App.tsx
mendo94: moved to the Landing directory, moved project portal to redux project slice, exported as consts

yeezick (approved):

PR #93, opened 2023-10-04
Feature/bc 579 delete account

Bestnogm1:
The user now has the ability to delete their account matched it with the figma


src/screens/Auth/Settings/DeleteAccountModal.tsx
```diff
@@ -0,0 +1,82 @@
...
+            <button
+              type='submit'
+              className='editprofile__cancelBtn'
+              onClick={handleClose}
+            >
+              Cancel
+            </button>
+            <button
+              className='delete-account-button'
+              onClick={() => confirmDeleteUsersAccount()}
```
yeezick, line 73: handlers without args should be used like `onClick={confirmDeleteUsersAccount}`

src/utils/api/users.ts
```diff
@@ -189,3 +189,12 @@ export const updateAvailability = async (
     return false
   }
 }
+
+export const deleteUsersAccount = async (id: any) => {
```
yeezick, line 193: function should be singular, we're only deleting 1 account

yeezick (changes requested):
tiny comments

yeezick (approved):

PR #94, opened 2023-10-11
fix alltickets bugs

yeezick:


yeezick:
temporary fix just to unbreak the pages. bug will actually be resolved with revamp create tickets - ticket

gooliajulia (approved):

PR #97, opened 2023-10-16
fixed the Log In Redirects User to Onboarding Flow: bug

Bestnogm1:
There wasn't really much of a bug here but i did notice a code that probably needs to be fixed 
`  const { project } = useAppSelector(selectAuthUser)` this is assuming that the user is in a project  and if they are not 
the  this functionality breaks,
`  const authUser = useAppSelector(selectAuthUser)` I recommend switching to this  and checking if the user has a valid 'projectId"


HectorIlarraza (approved):
Straight and clean, Approved!

PR #99, opened 2023-10-23
Feature/BC-600 Automatch user timezone

gooliajulia:
This PR implements to logic to guess a user's local timezone and auto-populate the timezone input for Onboarding flow with that zone.

It also includes the data and logic to transform DayJS, UTC and our UX defined user friendly timezone strings, and ensure's that once a user has chosen and saved their timezone, that timezone will populate in the Availability component.

It also includes a placeholder to eventually handle when a user's stored TZ does not match what we determine is their local timezone.


HectorIlarraza (approved):
PURE LOGIC, CLEAN AND APPROVED!!! 🧼

PR #103, opened 2023-11-12
Feature/BC-617 email subscription flow

gooliajulia:
We decided to scrap this flow for Beta launch, but will eventually want to implement it. This work was done before this decision was made, and will still be valuable done the line once we implement the email subscription flow.

https://bootcamper.atlassian.net/browse/BC-617

This PR:
* Builds the foundational styling and functionality of the Email Preferences component
* Sets up path for a user's email preferences screen
* Sets up 2 api utils for fetching preference options and current user preferences


yeezick:
implementation looks fine but i'm curious why we need to store email pref types in the backend and make a call to fetch them?

gooliajulia:
@yeezick The idea was to have one source of truth so we only have to change the information in one place. Not necessarily beneficial if this info only needs to be accessed by one place, but eventually might be used my more places in backend and frontend.

I'm okay with keeping it or nixing it and just storing on FE

src/screens/EmailPreferences/EmailPreferences.tsx
```diff
@@ -0,0 +1,88 @@
...
+        <div className='unsubscribe-all-banner'>
+          <div className='unsubscribe-all-banner-text'>
+            <h4>Unsubscribe from all emails.</h4>
+            <p>We'll still send account access emails.</p>
+          </div>
+          <Switch checked={disableAll} onChange={handleUnsubscribeSwitch} />
+        </div>
+        <div className='email-preferences-submit'>
+          <PrimaryButton
+            handler={() => console.log('save preferences')}
```
yeezick, line 80: console.log

src/screens/EmailPreferences/EmailPreferences.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const handleUnsubscribeSwitch = () => {
+    toggleDisableAll(!disableAll)
+  }
+
+  const handleChecks = (e, option) => {
+    setUserPreferences({
+      ...userPreferences,
+      [option]: !userPreferences[option],
+    })
+  }
```
yeezick, line 40: since you're passing option as the name you can destructure it from `e`, removing the need to pass a second argument and the anon function in the checkbox

yeezick (changes requested):
marking as requested changes

yeezick:
closing this PR but we know it exists if we need to revive it

PR #108, opened 2023-12-08
Feature/bc 488 show invitee availability

yeezick:
Accomplishes [BC-488: Add availability to meeting modal](https://bootcamper.atlassian.net/browse/BC-488?atlOrigin=eyJpIjoiZDRkODIzMTE3NzAwNDMwYzliMDQyYWYzYmMwMjVlMTEiLCJwIjoiaiJ9)
* Adds availability each member checkbox 
* Separates SelectAttendees components and stylesheets into their own files 
* Creates a directory for stylesheets


src/components/Calendar/MeetingModal/SelectAttendees.tsx
```diff
@@ -37,14 +34,13 @@ export const SelectAttendees = ({
...
           <FormGroup>
-            {filteredMembers.map(currMember => (
-              <div key={`select-member-${currMember._id}`}>
-                <MemberCheckbox
-                  attendees={attendees}
-                  currMember={currMember}
-                  setAttendees={setAttendees}
-                />
-              </div>
+            {projectMembers.map(currMember => (
```
HectorIlarraza, line 37: Should a key be added for better optimization?

src/interfaces/UserInterface.ts
```diff
@@ -132,49 +132,36 @@ export const timeOptions = [
 
 export type TimeOption = (typeof timeOptions)[0]
 
+//TODO: Removed string from sun interface....
```
HectorIlarraza, line 135: Seems like this can be removed now with type string gone no?

HectorIlarraza (changes requested):

src/components/Calendar/MeetingModal/SelectAttendees.tsx
yeezick: a key is provided in the member checkbox component itself

src/interfaces/UserInterface.ts
yeezick: yes actually

HectorIlarraza (approved):

PR #117, opened 2024-01-03
Feature/bc 620 user not matched with a team

HectorIlarraza:
### Feature: User Not Matched With a Team

This pull request involves implementing error handling for users who have not yet been added to any team within the Bootcampr platform. Additionally, it includes creating a designated screen for unassigned users to provide guidance and information.

### Changes Made:

1. Project Portal Access for Unassigned Users:

- Implemented error handling for users who have completed onboarding but have not been added to any team or when the team composition (3 SWDs and 2 UXDs) is incomplete.
- Enabled access to the project details section within the project portal for these users, allowing them to view project information.

2. Limited Access for Unassigned Users:

- Users without team assignments can access the team page but are restricted from accessing the calendar and task management sections.
- Inactivated the submit button in the side menu for users not yet assigned to a team.

3. Unassigned User Page Creation:

- Implemented an "Unassigned User" screen specifically for users not added to any team.
- Implemented a message within the Teams section informing users that Bootcampr is in the process of matching them to a team.

### Review Notes:

Implemented error-handling mechanisms for users not currently matched with a team or when the team composition requirements are incomplete. These users can access the project details section but cannot access other functionalities like calendar and task management.

Additionally, an "Unassigned User" screen has been successfully developed, displaying a message within the Teams section to inform users about the ongoing process of team assignment.

### Deployment Notes:

No specific deployment requirements or changes are necessary for this feature implementation. The updates will be included in the next deployment cycle.

**_Please review the changes and provide feedback as necessary. Thank you!_**


src/screens/Project/ProjectDetails/NotInTeamProjectAlert.scss
```diff
@@ -0,0 +1,21 @@
...
+    @include alert-banners(
+      $alert-warning-background,
+      $alert-warning-text-color,
+      100%
+    );
+
+    div,
+    a {
+      font-size: 16px;
+      font-size: 400;
```
GeoSauer, line 17: Is this one supposed to be `font-weight: 400`?

src/screens/Project/ProjectDetails/NotInTeamProjectAlert.tsx
```diff
@@ -0,0 +1,50 @@
...
+    status: false,
+    text: '',
+    type: '',
+  })
+  const { projectId } = useParams<{ projectId: string }>()
+
+  useEffect(() => {
+    const getProjectMembers = async () => {
+      try {
+        const response = await getProjectByUser(projectId)
```
GeoSauer, line 19: New to the product and what all the API routes do so bare with me, but should this be `getOneProject(projectId)` or `getProjectByUser(userId)`?

src/layout/SideMenu/SideMenu.scss
```diff
@@ -20,6 +20,23 @@
     line-height: 22px;
     width: 216px;
   }
+
+  .disable-btn {
+    align-self: center;
+    background-color: #fa9413;
+    border-radius: 4px;
+    border: none;
```
GeoSauer, line 28: I think these two need to be swapped for alphabetical-ness?

GeoSauer (changes requested):
Great work! I only noticed a couple small edits, and had one question due to my current unfamiliarity with the repos.

src/layout/SideMenu/SideMenu.tsx
```diff
@@ -13,47 +10,54 @@ import './SideMenu.scss'
...
+    notInTeam ? 'disabled-btn' : 'completion-overflow-btn'
+  }`
 
   return (
     <div className='sidemenu-container'>
       <div className='title'>
         <h2>{title}</h2>
       </div>
-      <SideMenuLinks />
+      <SideMenuLinks notInTeam={notInTeam} />
```
yeezick, line 24: no need to useAppSelector in this component if we are only drilling it down into sidemenulinks, can just call the selector within sidemenulinks

yeezick: ah i see, that we are using it for the button, regardless, dont make sidemenulinks dependent on the prop being passed; just call the selector again within that component

src/layout/SideMenu/SideMenu.tsx
```diff
@@ -13,47 +10,54 @@ import './SideMenu.scss'
...
       {links.map(link => (
-        <MenuLink linkDetails={link} />
+        <MenuLink linkDetails={link} notInTeam={notInTeam} />
       ))}
     </div>
   )
 }
 
-const MenuLink = ({ linkDetails }) => {
+const MenuLink = ({ linkDetails, notInTeam }) => {
```
yeezick, line 49: same as above comment

src/screens/Project/ProjectDetails/NotInTeamProjectAlert.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { GoAlert } from 'react-icons/go'
+import './NotInTeamProjectAlert.scss'
+
+export const NotInTeamProjectAlert = () => {
+  const [alertBanner, setAlertBanner] = useState<AlertBanners>({
+    status: false,
+    text: '',
+    type: '',
+  })
+  const { projectId } = useParams<{ projectId: string }>()
```
yeezick, line 14: convert { projectId: string } into a named type

src/screens/Project/ProjectDetails/NotInTeamProjectAlert.tsx
yeezick: why do we need to make an api call to check if project exists? 
can't we just check if the auth user has a project id within redux?

src/screens/Project/ProjectDetails/NotInTeamProjectAlert.tsx
```diff
@@ -0,0 +1,50 @@
...
+    getProjectMembers()
+  }, [])
+
+  return (
+    <>
+      {alertBanner.status ? (
+        <div className='not-in-team-alert'>
+          <div className={`nita-${alertBanner.type}`}>
+            {alertBanner.icon}
+            <div dangerouslySetInnerHTML={{ __html: alertBanner.text }} />
```
yeezick, line 44: dangerous, is this the only way to make this work?

src/screens/Project/ProjectDetails/NotInTeamProjectAlert.tsx
```diff
@@ -0,0 +1,50 @@
...
+          </div>
+        </div>
+      ) : null}
```
yeezick, line 47: no need for ternary if falsy condition will return null, just use &&

src/screens/Project/ProjectDetails/NotInTeamProjectAlert.tsx
HectorIlarraza: Well, I am calling the API to trigger the response message from the backend (see backend PR ticket) that was already set up in that way to display the alert for it in the frontend. I could use the redux approach but does that mean I need to create the message from the FE instead? This would help avoid using the "dangerouslySetInnerHTML" I had to use to get it to work with the backend message.

src/screens/Project/ProjectDetails/NotInTeamProjectAlert.tsx
HectorIlarraza: Its because the message is coming from the backend and how long the message is with the line breaks needed to space it out to match Figma is the only way i could get it to display correctly. But if i use the redux method discussed in the earlier comment above i can avoid this method. I will bring up in today's meeting.

src/utils/redux/slices/projectSlice.ts
```diff
@@ -149,6 +144,11 @@ const projectSlice = createSlice({
   },
 })
 
+export const selectUserNotInTeam = state => {
+  const { designers, engineers } = state.project.members
+  return designers.length === 0 && engineers.length === 0
+}
+
```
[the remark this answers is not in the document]
HectorIlarraza: I was going by at least having a team member or being a part of a team when i created this but if the projectID is good enough then I'll make that change to reflect that but that means i would also need to do that for the BE and FE conditional as well which looks for the member count to trigger the alert. Depending on removing the message from the BE have the FE handle it like i mentioned in the above comments.

src/screens/Project/ProjectDetails/NotInTeamProjectAlert.tsx
yeezick: If I'm understanding this correctly -- yes, let's avoid making an API call just for an error message.

src/utils/redux/slices/projectSlice.ts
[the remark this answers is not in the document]
yeezick: yes, let's rely solely on the projectId

yeezick (approved):

PR #124, opened 2024-01-12
Hifi/bc 640 hifi sign up and log in

GeoSauer:
This PR adds an error state and message, when fields are left blank, to First Name, Last Name, and Email input fields on loss of focus.

It also adds the `aria-required` attribute to all fields to assist screen readers.


HectorIlarraza (approved):
Great Work Geo, Approved!

PR #126, opened 2024-01-16
BC-638/HiFi Onboarding Flow

gooliajulia:
This PR:
* Matches complete Onboarding flow styling to Figma designs
* Replaces hardcoded buttons with BC component buttons
* Expands BC component buttons to allow for dynamic use: disabled, custom style tweaks (only intended for minor stylistic changes, styling will always default to the standard design system button)
* Cleans up data storage on Onboarding flow so user info is always stored and auto populated
* Fixes timezone input bugs
* Creates reusable BC Tool Tip component
* Ensures a user will be directed to the screen of Onboarding flow where they left off
* Adds error states and disable "next" button logic to entire flow
* Adds a real Google Forms survey for user's to click to and fill out
* Matches Paginator logic to Figma (cannot navigate manually and only highlights current page, not all completed) and adjusts Paginator component accordingly


src/components/Availability/Availability.scss
```diff
@@ -50,12 +57,12 @@
             flex-direction: row;
             height: fit-content;
             width: 100%;
-
+            //
```
yeezick, line 60: beep

src/components/Availability/subcomponents/TimeslotInput.tsx
```diff
@@ -40,6 +44,7 @@ export const TimeSlotInput = ({ day, days, setDays }) => {
       {consolidateAvailability(days[day].availability).map((slot, idx) => (
         <div key={`${slot}-${idx}`} className='timeslot-input'>
           <div className='left-banner'>
+            {/* TODO: make a reusable component for these? */}
```
yeezick, line 47: should this be followed up on?

src/components/Availability/subcomponents/TimeslotInput.tsx
yeezick, line 1: wonder if it would be worth to break into multiple files, not a hard block though

src/screens/Onboarding/Onboarding.scss
yeezick, line 1: ideally this stylesheet should be componentized, not a hard block, can be followed up

src/screens/Onboarding/Onboarding.tsx
```diff
@@ -4,8 +4,55 @@ import { Role } from './Role'
...
+
+    return (
+      firstName.length > 0 &&
+      lastName.length > 0 &&
+      bio &&
+      bio.length > 0 &&
+      links &&
+      links.linkedinUrl &&
+      links.linkedinUrl.length > 0
+    )
```
yeezick, line 53: do we need the checks for .length? 
`''` is treated as falsy in conditionals

src/screens/Onboarding/SetUpProfile.tsx
```diff
@@ -8,35 +8,61 @@ import { updateUser } from 'utils/api/users'
...
+      authUser.bio && authUser.bio.length > 0 ? authUser.bio.length : 0
+    setBioCharCount(charCount)
+
+    const { firstName, lastName, bio, links } = updateUserForm
+    const validForm =
+      firstName.length > 0 &&
+      lastName.length > 0 &&
+      bio.length > 0 &&
+      links.linkedinUrl &&
+      links.linkedinUrl.length > 0
```
yeezick, line 56: do we need to check for .lengths? 
`''` is treated as falsy in conditionals

src/screens/Onboarding/WhatsNext.tsx
```diff
@@ -23,24 +24,32 @@ export const WhatsNext = ({ handlePageNavigation }) => {
...
-            We love feedback. Please take <span> this short survey </span> so we
-            can improve.
+            We love feedback. Please take{' '}
+            <span>
+              {' '}
+              <a href='https://forms.gle/vfdAQpNMv2tZBwBB6' target='_blank'>
+                {' '}
+                this short survey
+              </a>
+            </span>{' '}
```
yeezick, line 41: linter dropped some `{' '}`

yeezick (changes requested):
linting changes, logic looks good, some suggestions to follow up on later but not hard blocks to this pr

src/components/Availability/subcomponents/TimeslotInput.tsx
gooliajulia: feels very much like a nice to have because the code repeats in this file - but since the component wouldn't be used outside of this case I think we can save the effort

src/screens/Onboarding/Onboarding.tsx
gooliajulia: Ahh good call. I must have been thinking of arrays. will update

src/components/Availability/subcomponents/TimeslotInput.tsx
gooliajulia: Just pulled CopyTimesModal into its own file

src/screens/Onboarding/Onboarding.scss
gooliajulia: Refactored

yeezick (approved):

PR #130, opened 2024-01-17
Hi fi/bc 657 chat

ozggnr:
- Updated general styling to match the Hifi,  removed unused files, functions
- added modals to edit the group chat name and errors/success toasts  
- fixed unread message badge 
- expand text field to max 6lines, added scrollbar  
- organized conversations' components and divided into different files
with those changes;
AC01, AC02, AC03, AC04, AC05 and AC08 [COMPLETED] 

AC06, AC07 and AC09 [NOT COMPLETED]
"Automated process of creating a group chat (team chat) and sending a message by bootcampr admin when a team is created"  [NOT COMPLETED]

BUG: If user receives a message while they are viewing the chat, it's marked as unread and message badge appears. (This issue will be resolved when we refactor the code)

[HiFi/bc-657-chat](https://bootcamper.atlassian.net/browse/BC-657)


src/components/ChatDialog/ChatDialogMain/ChatDialogMain.tsx
```diff
@@ -147,21 +157,84 @@ const ChatTitle = ({
       </div>
     ),
   }
-
+  const handleChangeChatName = async () => {
+    try {
+      setDisplayName(displayName)
```
yeezick, line 162: is it possible to rely solely on the redux displayName instead of using both with local state

src/components/ChatDialog/Messages/Messages.scss
```diff
@@ -1,46 +1,54 @@
...
+    // }
+
+    // &:hover::-webkit-scrollbar-thumb,
+    // &.scrolling::-webkit-scrollbar-thumb {
+    //   background-color: rgba(0, 0, 0, 0.2);
+    // }
+
+    // &.scrolling::-webkit-scrollbar-thumb {
+    //   opacity: 1;
+    // }
```
yeezick, line 39: if unused it's safe to delete

yeezick (approved):
i know this is a work in progress so it's safe to merge for now
one suggestion would be to convert length bits of codes into modular functions if possible, then move them to utils make the core logic easier to read

for example:
```
useEffect(() => {
const a = randomNumber(1)
const b = increaseByOne(a)
const c = multiplyByThree(b)
},[])
```

is easier to read than defining those functions all at once or spilling all the logic directly into the useEffect

src/components/ChatDialog/ChatDialogMain/ChatDialogMain.tsx
ozggnr: yes, i think, we can use redux for this one. I'll clean, refactor and split all these files.

src/components/ChatDialog/Messages/Messages.scss
ozggnr: Sure, I will delete it

PR #131, opened 2024-01-22
BUGFIX: BC-662 Profile View bug for users not assigned a team

gooliajulia:
This PR address the `Cannot destructure property firstName as it is undefined` Error message when trying to View Profile of a user not yet assigned a project by:
* Checking if the profile to view is for the current user before falling back on checking the teamMembers array


GeoSauer (approved):
Everything looks good, good work!

PR #135, opened 2024-01-27
Feature/bc 649 hifi calendar

jordanje:
- Delete functionality implemented for a singular event
- Organizer title added to event attendees
- Fixed avatars for meeting modal
- Styling changes to calendar


src/screens/Calendar/CalendarView.scss
```diff
@@ -1,14 +1,60 @@
+.css-19kzrtu {
+  padding: 0px;
```
gooliajulia, line 2: alphabetize style sheets 🙃

Check out the "CSS Alphabetize" extension on VSCode, it can auto-alphabetize the file for you. Just make sure to give it a quick look-over after, since the `&` character can can categorized a bit weird

src/screens/Calendar/MeetingModal/DisplayMeetingModal/DisplayAttendees.tsx
```diff
@@ -56,6 +64,9 @@ const InvitedMember = ({ member }) => {
         <p onClick={() => navigate(`/users/${member.userId}`)}>{memberName}</p>
         <p className='role'>{member.role}</p>
       </div>
+      {member.organizer === true ? (
+        <p className='organizer'>Organizer</p>
+      ) : null}
```
gooliajulia, line 69:
```suggestion
      {member.organizer && (
        <p className='organizer'>Organizer</p>
      )}
```

Can this be condensed to use the guard operator `&&` instead of a null ternary?

src/interfaces/CalendarInterfaces.tsx
```diff
@@ -5,6 +5,7 @@ import { MeetingModalInfo, ModalDisplayStatus } from './MeetingModalInterfaces'
 export interface Attendee {
   email: string
   responseStatus: string
+  comment: string
```
gooliajulia, line 8: It seems like this attribute is only used to determine if the Attendee is an organizer or not? Would it make more sense to name this attribute `isOrganizer` with value type boolean? (If there is potential for a third+ option beyond organizer or not, this works, but otherwise I think a boolean and name change is more fitting!)

gooliajulia (commented):
Awesome work! Pulled down the branch and the overall look and functionality is 🔥 

I left a couple comments on the PR and some other notes about matching the Figma below:
* 'Week <x>' between arrows at top left of Calendar is on Figma but not on your branch - Is this something you discussed with UX that was too difficult to include and UX agreed we could nix it?
* X to close out Create Meeting modal should be at far right side to match Figma
* Timezone options don't match Figma

src/interfaces/CalendarInterfaces.tsx
jordanje: So these keys match the ones that are provided by google events. For this one specifically, it only accepts a string. I see what you mean though...it is def not an ideal situation

gooliajulia: Ahh interesting, okay cool thanks for clarifying!

gooliajulia (approved):

PR #139, opened 2024-02-02
brand new commit for my chat hifi

ozggnr:
UPDATE
In addition to correcting the initial review, I added a few things that were missed earlier
- sorted names list in the edit chat room and create new chat room pages
- fixed socket to prevent multiple connection
- added hover state to edit pen and message send button

First commit
- Refactored and removed unused and cleaned up components
- matched styling figma hifi
- created useSocketEvent hook and added sockets to listen changes in related pages - when user is in the room (active) and when they are not in the room (inactive)
- calculated and displayed unread message count on navigation when they open the page
- created generic badge component to use on message icon and chat list
- made user thumbnail more generic to use in different pages
- expand message text box in chat room
- added invite members page with updated users list and send email notification
- removed projectByUser, which was removed from backend before
- created generic component for empty chat pages when there are no messages

[HIFI/BC-657-chat](https://bootcamper.atlassian.net/browse/BC-657)


src/components/ChatDialog/ChatDialogMain/ChatDialogMain.scss
yeezick, line 1: pls alphabetize stylesheets so css properties are easier to find
use this comment as a general comment for all scss files. 

also, add a line break between each selector

yeezick, line 1: pls alphabetize stylesheets so css properties are easier to find
use this comment as a general comment for all scss files.

also, add a line break between each selector

src/components/ChatDialog/ChatDialogMain/ChatDialogMain.tsx
yeezick, line 1: love this file's componentization

src/components/ChatDialog/ChatDialogMain/ChatHeader.tsx
```diff
@@ -0,0 +1,151 @@
...
+    setDisplayName(value)
+  }
+  const handleCancel = () => {
+    setOpenEditNameModal(false)
+    setDisplayName('')
+  }
+  const profilePictures = extractConversationAvatars(
+    currentConversation.participants,
+    authUser._id
+  )
```
yeezick, line 109: add line breaks between functions

yeezick: use as general comment for all functions

src/components/ChatDialog/ChatDialogMain/ChatHeader.tsx
```diff
@@ -0,0 +1,151 @@
...
+  return (
+    <div className='page-title'>
+      <FiArrowLeft size={24} onClick={handleBackArrowClick} />
+
```
yeezick, line 113: delete white space

yeezick: use as general comment

src/components/ChatDialog/ChatDialogMain/ChatHeader.tsx
```diff
@@ -0,0 +1,151 @@
...
+          avatarSize={'small'}
+          avatarType={
+            currentConversation.chatType === 'group' ? 'grid' : 'single'
+          }
+        />
+      )}
+      <div className='title-with-icon'>
+        <h5
+          onClick={() => dispatch(onScreenUpdate(ChatScreen.EditChatRoom))}
+          className='group-link'
```
yeezick, line 126: lets name this function above so it is easier to debug 
anon functions waste memory

yeezick: use as general comment

src/components/ChatDialog/ChatMemberProfile/ChatMemberProfile.tsx
yeezick, line 1: does this file need a follow up?

src/components/ChatDialog/ChatRoom/ChatRoom.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const handleTimestampClick = (message: any) => {
+    // Logic to display/hide timestamp on message click:
+    // Check if selected message is already present in selectedMessages array
+    if (selectedMessages.some(msg => msg === message)) {
+      // If it exists, remove it from the array
+      setSelectedMessages(selectedMessages.filter(msg => msg !== message))
+    } else {
+      // If it doesn't exist, add it to the array
+      setSelectedMessages([...selectedMessages, message])
+    }
```
yeezick, line 36: remove extra comments, only one worth keeping is the one on line 28

src/components/ChatDialog/ChatRoom/ChatRoom.tsx
```diff
@@ -0,0 +1,150 @@
...
+      sendMessage(newMessage)
+
+      dispatch(setChatText(''))
+    } catch (error) {
+      console.error('Error sending message:', error)
+    }
+  }
+
+  const handleKeyDown = e => {
+    // Send message with 'Enter' key, Does not send with 'Enter' + 'Shift'
```
yeezick, line 71: i actually think we should prefer to send with shift enter over just shift.. we can bring up with UX if needed

src/components/ChatDialog/UserThumbnail/UserThumbnail.tsx
yeezick, line 1: we do have a teamAvatar component which uses just a user Id to generate a user image and might help to reduce the number of required props here. 
look into it to see if it can be used here instead, if not, this is fine to keep

src/components/Notifications/Socket.tsx
```diff
@@ -1,20 +1,143 @@
...
+//     // const askUserPermission = async () => {
+//     //   return await Notification.requestPermission()
+//     // }
+//     // askUserPermission()
+
+//     setSocket(io(ENDPOINT, { transports: ['websocket'] }))
+//   }, [])
+
+//   return socket
+// }
```
yeezick, line 29: if this isn't needed we can remove it and always reference this pr instead of keeping dead code in the code

src/components/Notifications/Socket.tsx
yeezick, line 1: you are now the socket SME 😅

src/interfaces/ChatInterface.ts
yeezick, line 1: interfaces should also be generally alphabetized
alphabetization makes it easier for us to locate properties we need instead of scanning the whole object for the key we are looking for

yeezick: use as a general comment

src/layout/Nav.tsx
```diff
@@ -6,56 +6,39 @@ import { MdArrowDropDown } from 'react-icons/md'
...
 import {
   selectChatUI,
+  selectUnreadMessages,
+  setChatRoomActive,
+  setUnreadChatsCount,
   toggleChat,
   toggleChatClose,
 } from 'utils/redux/slices/chatSlice'
-import { ChatIconBadge } from 'components/ChatDialog/ChatIconBadge/ChatIconBadge'
+
```
yeezick, line 19: remove whitespace
use as a general comment

src/layout/Nav.tsx
```diff
@@ -109,12 +92,19 @@ export const Nav = () => {
...
   const chatRef = useRef(null)
-  const dispatch = useAppDispatch()
+  useSocketEvents(false)
+  useEffect(() => {
+    const fetchUnreadCount = async () => {
+      const unreadMessages = await getUnreadChatMessageCount()
+      dispatch(setUnreadChatsCount(unreadMessages.count))
+    }
+    fetchUnreadCount()
+    dispatch(setChatRoomActive(false))
```
yeezick, line 106: i feel like unread count should be determined on the frontend after we fetch all chat messages on load

don't want to make unnecessary api calls but platforms like amplify charge us for each outbound/inbound api request we make

src/utils/functions/chatLogic.js
```diff
@@ -1,17 +1,7 @@
...
-  const month = date.getMonth() + 1
-  const day = ('0' + date.getDate()).slice(-2)
-  let hours = date.getHours()
-  hours = hours % 12
-  hours = hours ? hours : 12
-  const minutes = ('0' + date.getMinutes()).slice(-2)
-  const ampm = hours >= 12 ? 'PM' : 'AM'
-  const formattedTime = `${hours}:${minutes} ${ampm}`
-  return `${month}/${day}/${year}, ${formattedTime}`
+  return dayjs(timestamp).format('MM/DD/YYYY hh:mm A')
```
yeezick, line 3: import `generateDayJs` instead of dayjs directly so we only have one instance of it that is loaded with all necessary plugins

yeezick (changes requested):
one word: impressive

very nice pr ozge. most comments are regarding linting or best practice, no real qualms with any of the logic other than what i have called out 

for comments with "use as general comment" please apply those comments to the whole PR

yeezick:
also, please adhere to branch practices for feature branches in future
for new features it should be `feature/BC-ID-desc`

src/components/ChatDialog/ChatMemberProfile/ChatMemberProfile.tsx
ozggnr: We don't have this functionality in the beta hifi and but we need to implement member's page in post soft launch. I prefer to remove it completely for now, we'll need  a very detailed members page for post launch so this might not be useful

src/components/ChatDialog/ChatRoom/ChatRoom.tsx
ozggnr: so if user hits enter, message will be sent, if user hits shift + enter, they will go to the next line. I like this approach, also I believe it's very common but we can discuss absolutely

src/components/ChatDialog/UserThumbnail/UserThumbnail.tsx
ozggnr: Oh actually this is something I want to discuss about, I believe we can make this one component since we're using this in different places like calendar, chat, team. this component includes avatar (can be team or single), name and description.

src/components/Notifications/Socket.tsx
ozggnr: It was painful but I love it in the end 😁

src/layout/Nav.tsx
ozggnr: Right now, we don't fetch messages on load, we load chat messages, when user enters a chat room, but unread message field belongs to participant anyway. Let me think about this more how we can handle differently

src/utils/functions/chatLogic.js
```diff
@@ -1,15 +1,7 @@
...
-  hours = hours % 12
-  hours = hours ? hours : 12
-  const minutes = ('0' + date.getMinutes()).slice(-2)
-  const ampm = hours >= 12 ? 'PM' : 'AM'
-  const formattedTime = `${hours}:${minutes} ${ampm}`
-  return `${month}/${day}/${year}, ${formattedTime}`
+  return generateDayJs(timestamp).format('MM/DD/YYYY hh:mm A')
 }
 
 export const formatLastMessageTimestamp = lastMessageTimestamp => {
```
yeezick, line 7: is this function still needed? if not delete it, if yes, can we apply format using dayjs as well?

src/utils/redux/slices/chatSlice.ts
```diff
@@ -62,49 +174,156 @@ const chatSlice = createSlice({
...
+    const copyThreads = [...threads]
+    return copyThreads.sort((a, b) => {
+      // Default dates for threads without a lastMessage
+      const defaultDate = new Date(0)
+      const dateA = a.lastMessage
+        ? dayjs(a.lastMessage.timestamp)
+        : dayjs(defaultDate)
+      const dateB = b.lastMessage
+        ? dayjs(b.lastMessage.timestamp)
+        : dayjs(defaultDate)
```
yeezick, line 296: using the vanilla dayjs import instead of generateDayJs creating unnecessary extra instances of it

src/components/ChatDialog/EditChatRoom/EditChatRoom.tsx
```diff
@@ -1,145 +1,66 @@
...
+        <div className='members-list'>
+          {participants.map(pp => (
+            <div className='member'>
+              <UserThumbnail
+                title={`${pp.participant.firstName} ${pp.participant.lastName}`}
+                description={pp.participant.role}
+                profilePicture={pp.participant.profilePicture}
+                avatarSize='medium'
+                avatarType='single'
+              />
```
yeezick, line 38: confused about this map methood, how are we iterating over each participant and then still have to go go into `pp.participant` to get their information? 

seems like its unnecessarily nested?

src/components/ChatDialog/ChatDialogMain/ChatDialogMain.scss
yeezick: this comment doesn't look like it was addressed... it should ideally also be applied to scss properties.

yeezick (changes requested):

src/components/ChatDialog/EditChatRoom/EditChatRoom.tsx
ozggnr: it's because of participants model
`participants: [
      {
        participant: { type: mongoose.Types.ObjectId, ref: 'User' },
        isAdmin: { type: Boolean, default: false, required: true },
        hasUnreadMessage: { type: Boolean, default: false },
      },
    ]`
 we can change the naming if it's confusing

src/utils/functions/chatLogic.js
ozggnr: Actually I missed that, thanks for reminding. This wasn't correct so i created a new one, and also i created dateFormatHelper function to hold these format functions. Let me know if that's ok or need to be moved to another place

src/components/ChatDialog/ChatMemberProfile/ChatMemberProfile.tsx
yeezick: if that's the case then delete this.

src/components/ChatDialog/ChatRoom/ChatRoom.tsx
yeezick: Unresolved

PR #143, opened 2024-02-13
refactor/BC-696-enable-project-completion-flow

GeoSauer:
This PR adds projectSubmissionDate to the timeline prop of the Project object.
I've used this prop to determine if the CTA that triggers the project completion flow is disabled.
I added an interval checker to evaluate whether the button is disabled that checks every minute, we can adjust if needed.


src/layout/SideMenu.tsx
```diff
@@ -6,13 +6,34 @@ import { changePortalPage } from 'utils/helpers'
...
+    +projectSubmissionDate > Date.now()
+  )
+
   const handleProjectCompletion = () =>
     navigate(`/project/${projectId}/complete`)
+
+  //TODO: Currently set to check every minute but we can adjust as needed
+  useEffect(() => {
+    const dateCheckInterval = setInterval(() => {
+      setIsDisabled(+projectSubmissionDate > Date.now())
```
ozggnr, line 31: could you use day.js library for the dates, I think we don't use native js dates anymore

ozggnr (changes requested):
Looks good Geo!

src/layout/SideMenu.tsx
GeoSauer: Probably yes? Does it return a Unix timestamp? Currently `startDate` and `endDate` are Unix timestamps so I just used this method to be in the same format. 

I may be basing this work off of old work because I do remember hearing conversation about day.js

PR #145, opened 2024-02-15
Refactor/bc 711 project portal language

GeoSauer:
This makes language adjustments across the project portal, including:

Project Details --> Product Details

Calendar --> Scrum Calendar

Task Management/Tasks --> Kanban Board/Stories

[BC-711](https://bootcamper.atlassian.net/browse/BC-711?atlOrigin=eyJpIjoiNjA0ZDdhZTk4ZjFmNDllNDg5MWFmNjBlMmY2OWNkNjUiLCJwIjoiaiJ9)

[BC-711]: https://bootcamper.atlassian.net/browse/BC-711?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/screens/Project/TaskManagement/TaskBoard/BoardColumns/ColumnTickets.tsx
```diff
@@ -4,14 +4,15 @@ import { TicketTab } from './TicketTab'
...
+
   return (
     <div
       className='tickets'
       {...provided.droppableProps}
       ref={provided.innerRef}
     >
       {visibleTickets[columnStatus].map((ticketDetail, idx) => (
-        <TicketTab idx={idx} ticketDetail={ticketDetail} />
+        <TicketTab key={ticketDetail} idx={idx} ticketDetail={ticketDetail} />
```
ozggnr, line 15: just curiosity, why don't we use id instead of ticketDetail for the key value?

ozggnr (commented):
Looks great Geo, I just left a question to learn why you used 'columnStatus' or 'ticketDetail' for key values and I'm wondering if there is a possibility that those values might be duplicated

src/screens/Project/TaskManagement/TaskBoard/BoardColumns/ColumnTickets.tsx
GeoSauer: Unfortunately the objects being mapped don't have an id, or any unique identifiers, so I just used the whole object for the key.

GeoSauer: Ya know what I went ahead and implemented a different key that should work better, good callout!

src/screens/Project/TaskManagement/TaskBoard/BoardColumns/ColumnTickets.tsx
```diff
@@ -4,14 +4,19 @@ import { TicketTab } from './TicketTab'
...
   return (
     <div
       className='tickets'
       {...provided.droppableProps}
       ref={provided.innerRef}
     >
       {visibleTickets[columnStatus].map((ticketDetail, idx) => (
-        <TicketTab idx={idx} ticketDetail={ticketDetail} />
+        <TicketTab
+          key={`${ticketDetail}+${idx}`}
```
ozggnr, line 16: I would suggest that instead of using whole object "ticketDetail" , we can use other fields and combine them like string date, createdAt, project id , user id, idx etc.. i think in this case "ticketDetal" is just [object Object]. I'm not sure if react can handle this.

GeoSauer: Third time's a charm! I'm not sure where I was looking last week but ticketDetail totally has an _id prop so that's the key now for TicketTab. For StatusColumn we're just mapping through an array of 4 strings so I'm using the string as the key. This wasn't even part of the ticket I just wanted to clean up the console 😅

ozggnr (approved):

src/screens/Project/TaskManagement/TaskBoard/BoardColumns/ColumnTickets.tsx
ozggnr: Sounds perfect Geo, thanks 🙌

PR #146, opened 2024-02-15
removes unused CalendarHeader component

GeoSauer:
exactly what it says on the tin


gooliajulia (approved):
yay for cleaning up dead code! thanks for calling this out and taking care of it

PR #147, opened 2024-02-16
Refactor/bc 720 display user email on congrats page

GeoSauer:
This PR accomplishes:

-  language on EmailSentConfirmationPage to adjusted to match Figma 
- adds unverifiedEmail prop to user as there info isn't stored until verification
- email address used to sign up is displayed
- button is present to activate update password flow but currently resends verification email, [BC-753](https://bootcamper.atlassian.net/jira/software/c/projects/BC/boards/2?selectedIssue=BC-753) will finish this off

[BC-720](https://bootcamper.atlassian.net/browse/BC-720?atlOrigin=eyJpIjoiNGM4NGU2MjJkM2U1NDA5ZmIzMmFkNzljZDMzNGU0MzkiLCJwIjoiaiJ9)

[BC-753]: https://bootcamper.atlassian.net/browse/BC-753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BC-720]: https://bootcamper.atlassian.net/browse/BC-720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/screens/Auth/SignUp/EmailSentConfirmation/EmailSentConfirmation.tsx
```diff
@@ -1,14 +1,30 @@
...
+      try {
+        const { email } = await getOneUser(newUserId)
+        setUserEmail(email)
+      } catch (error) {
+        console.error('Error fetching user data:', error)
+      }
+    }
+
+    fetchUserEmail()
+  }, [newUserId])
```
yeezick, line 25: do we need to make an api call to get the user's email? their information should all be stored in redux already after they begin the onboarding flow

want to minimize the amount of api calls make as cloud services like aws charge us for # of requests and size of  requests, inbound and outbound

GeoSauer: So this is the product of a weird series of events, and may be due to my new exposure to redux. From what I could tell the user isn't fully stored until after they confirm their email? 

When trying to pull user info from that page it was returning undefined, and if you check out the commit where I added the `unverifiedEmail` email prop you can see that I added that prop and the means to access it. It worked and displayed properly but did not persist on refresh.

GeoSauer: Maybe the issue is that most things revolve around an authUser currently and they aren't signed in at this point?

yeezick: ahh i see. i thought they were immediately logged in afterword but i see the issue now. 

maybe another workaround is to dispatch the user's email to redux after they click the sign up button? but to only store the user email at that point 

this way we can still save ourselves from the api call

yeezick (changes requested):
this is good to go after my response to the last comment

src/screens/Auth/SignUp/EmailSentConfirmation/EmailSentConfirmation.tsx
GeoSauer: Is that something we want to add to this ticket or will it be its own?

yeezick: i think it should be in this ticket as it directly relates to it

basically just replacing the useEffect from the email confirmation page with a selector for the user's email & dispatching the call to update the user slice on success response from the sign in controller

yeezick (approved):

PR #152, opened 2024-02-19
Feature/bc 746 adding week and year

jordanje:
- Added the text "Week" in between prev and next buttons in calendar header
- Added month, year in the calendar header


src/screens/Calendar/CalendarView.tsx
```diff
@@ -70,23 +70,29 @@ export const CalendarView = () => {
...
+            titleFormat={{ year: 'numeric', month: 'long' }}
             allDaySlot={false}
             initialView='timeGridWeek'
             nowIndicator={true}
             plugins={[dayGridPlugin, timeGridPlugin]}
-            validRange={{
-              start: firstDay,
-              end: lastDay,
-            }}
+            validRange={
```
jordanje, line 82: These comments will be updated in my next story this sprint. Currently just acting as a placeholder

PR #153, opened 2024-02-20
Stability/Update versioning GitHub action

gooliajulia:
The version bump github action is running into "already exists" errors on the newly created tags since it runs on the opening of each pull request. Instead it should only run when a PR is ready to merge to main to avoid this race condition with PRs being open at the same time


GeoSauer (approved):
Good catch, hope this works!

PR #159, opened 2024-02-23
memoized userids to solve performance issue and updated project slice

ozggnr:
- memoized userids to solve the redux performance issue for selectMembersById selector and selector is also updated accordingly


src/screens/Project/TaskManagement/TicketDialog/Fields/TicketCreator.tsx
```diff
@@ -7,7 +7,10 @@ import { TeamAvatar } from 'components/TeamAvatar/TeamAvatar'
 
 export const TicketCreator = () => {
   const { createdBy } = useAppSelector(selectTicketFields)
-  const [user] = useAppSelector(selectMembersById([createdBy]))
+  const memoizedCreatedBy = useMemo(() => [createdBy], [createdBy])
+  const [user] = useAppSelector(state =>
+    selectMembersById(state, memoizedCreatedBy)
+  )
```
yeezick, line 13: i see we created the function for select members by id but this anon function should also be converted into a named function since it's likely to be repreated

yeezick (approved):

PR #162, opened 2024-02-26
establish skeeleton stripe checkout button for payments

yeezick:
sets up skeleton integration for stripe payments


GeoSauer (approved):
Looks solid!

PR #163, opened 2024-02-26
Feature/bc 746 adding week and year

jordanje:
In this PR:
- Added dynamic "week" text in between prev/next buttons in calendar header
- Added Month, Year title in calendar header
-  Uses Calendar API to handle dynamic week text


src/screens/Calendar/CalendarView.tsx
```diff
@@ -27,11 +31,13 @@ export const CalendarView = () => {
...
   const [eventFetchingStatus, setEventFetchingStatus] = useState('loading')
   const timeline = useAppSelector(selectProjectTimeline)
-  //use startDate and endDate for validRange in FullCalendar
+  const [weekNumber, setWeekNumber] = useState('')
+
   dayjs.extend(weekday)
-  const firstDay = dayjs(timeline.startDate).weekday(0).format('YYYY-MM-DD')
-  const lastDay = dayjs(timeline.endDate).weekday(7).format('YYYY-MM-DD')
 
+  const firstDay = timeline.startDate
```
GeoSauer, line 38: Why don't you need to do any dayjs shenanigans here? Is it because you're passing it into dayjs in the updateWeekNumber function in calendarHelpers?

GeoSauer (approved):
Everything looks great! I haven't messed with dayjs much yet so it's cool to see it implemented.

PR #165, opened 2024-02-27
Stability/Update URL format

gooliajulia:
Heroku adds a unique identifier to domains for each app. (See ref in [Custom Domain Names](https://devcenter.heroku.com/articles/custom-domains#:~:text=Heroku%20started%20appending%20identifiers%20to%20subdomains%20on%20June%2014%2C%202023.%20Applications%20created%20before%20this%20date%20follow%20the%20old%20subdomain%20scheme%20where%20the%20subdomain%20was%20the%20same%20as%20the%20app%20name.) )

"Heroku started appending identifiers to subdomains on June 14, 2023. Applications created before this date follow the old subdomain scheme where the subdomain was the same as the app name."

... so our URL constructor and environment variables need to account for this subdomain identifier.

Once these changes are approved and merged, I'll add the corresponding identifier via  `REACT_APP_API_SUBDOMAIN` environment variable to each Heroku app environment


HectorIlarraza (approved):
Approved!

PR #172, opened 2024-03-10
Feature/bc 629 auth layer

ozggnr:
- Removed the auth layer from the Layout component and implemented an AuthWrapper. This ensures unauthorized users are prevented from accessing certain pages, redirecting them to the sign-in page instead.
- Created EmailRouter component to re-route users to the related pages when they click on link on the notification email
- Implemented a chatRouterHandler to route users to the project page without logging in if they logged in before (if we still have the auth token), along with a feature to toggle the chat box. This enables the display of the chat room to which users have been invited
- Removed unnecessary code to prevent bugs
- Added very simple 404 for unmatched routes

NOTE: when we change the id in the "project/id" route, it still displays this page. what's the plan for this type of case? if we still want to display 404 we need to deal with it differently.


yeezick (approved):
all looks good to me, good work ozge!

yeezick (approved):

PR #173, opened 2024-03-13
Feature/bc 783 develop payments flow

yeezick:
Accomplishes [BC-783: develop payments flow](https://bootcamper.atlassian.net/jira/software/c/projects/BC/boards/2?selectedIssue=BC-783)
* styles and implements the "choose experience" screen
* styles and implements the "max users" screen
* adds the payments interface to the user modal and interface
* removes logic to auto-sign-in the user after email verification
* adds the payment bar to the onboarding paginator

Missing functionality for when a user decides to go the "sandbox" experience but that is expected to be followed up on with the freemium implementation tickets


src/components/AccountDropdown.tsx/AccountDropdown.tsx
```diff
@@ -37,7 +37,7 @@ export const AccountDropdown = ({ anchorEl, closeDropdown }) => {
       dispatch(logoutAuthUser())
       navigate('/')
     } else if (innerText === 'Payment') {
-      navigate('/checkout')
+      navigate('/payment/checkout')
```
ozggnr, line 40: I might have missed this but are we planning to implement the checkout page later, or does this route already exist?

ozggnr (changes requested):

src/components/AccountDropdown.tsx/AccountDropdown.tsx
yeezick: I had this checkout page just a placeholder to test the stripe integration 

Instead, the button to complete payment will call an api which will return a stripe-hosted URL and automatically redirect the user to stripe to complete checkout. 

After payment, the user will be rerouted to the success screen. 

Basically we do not own the checkout screen at this time

ozggnr (approved):

ozggnr (approved):

PR #182, opened 2024-03-22
Feature/bc 784 updated project prompt

jordanje:
in this PR:
- I updated the Overview page to be formatted as an accordion using MUI
- Added additional text from the figma as well as the product requirements link
- Styled the accordion to match the figma


src/screens/Project/ProjectDetails/Overview.scss
```diff
@@ -77,4 +72,32 @@
       margin: 0;
     }
   }
+
+  .css-1c35hjw-MuiPaper-root-MuiAccordion-root.Mui-expanded {
```
yeezick, line 76: dont use these selectors to style buttons because the `.css-[ALPHANUMERIC CODE]` is randomly generated  so it wont always work 

the .Mui-paper-etc selectors should be enough but sometimes its tricky to find the right selector for them. 

i've learned a few tricks on how to do this so feel free to reach out to me if you need help figuring it out.

src/screens/Project/ProjectDetails/Overview.tsx
yeezick, line 1: is it possible to turn each accordion into a component and/or a reusable parent accordion with dynamic content?

yeezick (approved):

GeoSauer (dismissed):
Looks good!

HectorIlarraza (approved):
🧼 Approved!

PR #192, opened 2024-04-02
feature/bc-804-updating-role-cards

jordanje:
In this pr I updated the images in the role cards, increased the size of the cards and the text, and updated the text to match the figma design


GeoSauer (approved):
Looks good!

PR #193, opened 2024-04-04
Bc 730 validate linkedin profile url

jordanje:
In this PR:
- I added an additional error state to Linkedin input to check if URL matches Linkedin URL format using RegExp
- Error state should trigger when the input isn't empty and when the input is invalid
- The error state will go away after after typing a correct linkedin URL in the input
- The "Complete Payment" button will remain disabled if isInvalidURL is true


HectorIlarraza (dismissed):
🧼 Approved

HectorIlarraza (approved):
Approved! 🧼

PR #194, opened 2024-04-04
Fix team and profile page crash

GeoSauer:
Adds some more null checks to selectMembersById function to prevent crashes


HectorIlarraza (approved):
Approved!

PR #195, opened 2024-04-04
Refactor/bc-735 hide gihtub input for non swes

GeoSauer:
On the profile page of the onboarding flow the GitHub URL input will only display if the user has selected Software Engineer as their role. 

On the View Profile screen the icon and header for each URL only shows up if the user has supplied that specific URL.


HectorIlarraza (approved):
🧼  Approved

PR #197, opened 2024-04-09
Refactor/bc 810 update onboarding endpoint

ozggnr:
[Remove userId from onboarding endpoint](https://bootcamper.atlassian.net/browse/BC-810)

- removed userId from onboarding endpoint
- removed user id from updateUserProfile since we're getting it from token now
- changed params id with authuser id


HectorIlarraza (approved):
approved! 🧼

PR #203, opened 2024-04-13
updates "Contact us" to "Contact Us"

mckubisiak:


src/utils/data/timelineConstants.tsx
```diff
@@ -18,7 +18,7 @@ export const timelineData = [
         <p>Ordering flow</p>
         <p>User acount log in</p>
         <p>Reservation flow</p>
-        <p>Contact us</p>
+        <p>Contact Us</p>
```
gooliajulia, line 21: Looks like this "Contact us" actually has the 'u' lowercase in the Figma

Could you also while you're here fix the typo on line 19 above from "User acount log in" to "User account log in"? 🙏 

Screenshot 2024-04-13 at 9 55 59 AM

gooliajulia: @mckubisiak 
One more, can you keep the 'u' lowercase on line 21 like the attached photo? It looks like only the nav bar "Contact Us" is uppercase

gooliajulia (approved):
looks good! ty

PR #214, opened 2024-04-22
Bugfix/bc-788 timezone not defaulting

gooliajulia:
In working on Availability bugs and refactoring, I noticed the timezone component wasn't managing state properly which led to controlled / uncontrolled component errors and the timezone input rendering blank rather than the guessed or stored user timezone.

This PR moves the state management of the timezone out of the parent components and into the Timezone Input Banner itself.

This resolves this bug as well: https://bootcamper.atlassian.net/browse/BC-788

### DIFF
--- a/src/utils/helpers/availabilityHelpers.ts
+++ b/src/utils/helpers/availabilityHelpers.ts
@@ -23,11 +23,11 @@ export const guessUserTimezone = () => {
     dayJSformattedTZdata[userTZ].utc
   ) {
     const utc = dayJSformattedTZdata[userTZ].utc
-    const userFriendlyTZ = utcToBootcamprTimezoneMap[utc]
+    const userFriendly = utcToBootcamprTimezoneMap[utc]
 
     return {
       utc,
-      userFriendlyTZ,
+      userFriendly,
     }
   } else {
     return undefined
--- a/src/components/Availability/Availability.tsx
+++ b/src/components/Availability/Availability.tsx
@@ -1,4 +1,4 @@
-import { useEffect } from 'react'
+import { useEffect, useState } from 'react'
 import { Timezones, weekdaysMap } from './utils/data'
 import {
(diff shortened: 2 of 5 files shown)


src/utils/helpers/availabilityHelpers.ts
```diff
@@ -23,11 +23,11 @@ export const guessUserTimezone = () => {
     dayJSformattedTZdata[userTZ].utc
   ) {
     const utc = dayJSformattedTZdata[userTZ].utc
-    const userFriendlyTZ = utcToBootcamprTimezoneMap[utc]
+    const userFriendly = utcToBootcamprTimezoneMap[utc]
```
GeoSauer, line 26: What was the reasoning behind removing the 'TZ' from this variable name?

GeoSauer (commented):
I just have one quick question regarding a change to a variable name but other than that everything looks good!

src/utils/helpers/availabilityHelpers.ts
gooliajulia: Good question. The code that calls this function is calling the parent object `guessedUserTimezone` so I felt the TZ suffix here was redundant!

code referenced here: https://github.com/yeezick/bootcampr/pull/214/files/73e515571327f867776a5bde185aba6b7c5fa6ea#diff-39055c20e7b19c42478987a94eb0eefcc2b04159d7aa4c90931ac5c3d1b427aaR19

GeoSauer: Sounds good, my knee jerk was that it made it less descriptive, but you're correct, it is redundant. Approved!

GeoSauer (approved):

src/utils/helpers/availabilityHelpers.ts
gooliajulia: good looking out

PR #221, opened 2024-04-26
bugfix/BC-800-profile-picture-upload-on-profile-setup-page 

HectorIlarraza:
narrow down error to redux or handleImageUpload func and handleFileIn…putChange func in profilePreviewImage file


src/screens/ProfilePreviewImage/ProfilePreviewImage.tsx
```diff
@@ -62,6 +63,7 @@ export const ProfilePreviewImage: React.FC<ProfilePreviewImageProps> = ({
    */
   const handleImageUpload = async (image: string) => {
```
HectorIlarraza, line 64: This function doesnt seem to hit but could be because it never gets the chance since it fails before by the Redux not passing the ID 
Screenshot 2024-04-26 at 5 15 53 PM

HectorIlarraza (commented):
HandleSave func could be a hint to solving the bug as well in the ImageEditorModal.tsx haven't checked it fully yet

PR #222, opened 2024-04-27
POC-Google Analytics

gooliajulia:
Adds Google Analytics "Gtag" to our site to set us up for Google Analytics data


public/index.html
```diff
@@ -35,6 +35,20 @@
...
+  <script
+    async
+    src="https://www.googletagmanager.com/gtag/js?id=G-0R4CX17X3C"
+  ></script>
+  <script>
+    window.dataLayer = window.dataLayer || []
+    function gtag() {
+      dataLayer.push(arguments)
+    }
+    gtag('js', new Date())
```
yeezick, line 48: should we sue the JS Date object or dayjs instead?

gooliajulia: Good catch. just pushed up the change to use dayjs instead

yeezick (approved):

PR #236, opened 2024-05-02
bug-fix sign up 'when filling about me 500/500 doesn't change'

mckubisiak:
updated useEffect to fix 'about me' character counter and removed redundant linkedin validation

### DIFF
--- a/src/screens/Onboarding/SetUpProfile.tsx
+++ b/src/screens/Onboarding/SetUpProfile.tsx
@@ -47,16 +47,12 @@ export const SetUpProfile = ({ handlePageNavigation }) => {
       })
     }
 
-    isLinkedInUrl(links.linkedinUrl)
-      ? setIsInvalidURL(false)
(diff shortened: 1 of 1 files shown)


src/screens/Onboarding/SetUpProfile.tsx
```diff
@@ -47,16 +47,12 @@ export const SetUpProfile = ({ handlePageNavigation }) => {
       })
     }
 
-    isLinkedInUrl(links.linkedinUrl)
-      ? setIsInvalidURL(false)
-      : setIsInvalidURL(true)
+    if (authUser && authUser.bio) {
+      setBioCharCount(authUser.bio.length)
+    }
```
GeoSauer, line 52: This doesn't appear to be necessary because `bioCharCount` is already being set in `handleInputChange`. I believe when you removed `const charCount = ...` from the `useEffect` after this you should have solved the issue.

